PR #3, opened 2016-06-17
adding an async function called "shout"

mapsam:
Adds an asynchronous function called `shout`. Here's what it looks like on the JS side:

```javascript
var HW = new HelloWorld();
HW.shout('hey', { louder: true }, function(err, shout) {
  if (err) throw err;
  console.log(shout); // => hey!
});
```

Next up:

- [x] handle different options
- [x] ~~return errors instead of throwing~~ #4 

cc @springmeyer @GretaCB

### DIFF
--- a/src/hello_world.cpp
+++ b/src/hello_world.cpp
@@ -72,15 +72,16 @@ NAN_METHOD(HelloWorld::wave)
 
 // this is the cpp object that will be passed around in our method and callbacks
 // referred to as a "baton"
-typedef struct {
+class AsyncBaton
+{
+  public:
     uv_work_t request; // required
     Nan::Persistent<v8::Function> cb; // callback function type
     std::string phrase;
-    // std::string format;
     bool louder;
     std::string error_name;
     std::string result;
-} helloworld_shout_baton;
+};
 
 NAN_METHOD(HelloWorld::shout)
 {
@@ -123,7 +124,7 @@ NAN_METHOD(HelloWorld::shout)
     v8::Local<v8::Value> callback = info[2];
 
     // set up the baton to pass into our threadpool
-    helloworld_shout_baton *baton = new helloworld_shout_baton();
+    AsyncBaton *baton = new AsyncBaton();
     baton->request.data = baton;
     baton->phrase = phrase;
     baton->louder = louder;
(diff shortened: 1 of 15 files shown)


test/hello_world.test.js
```diff
@@ -1,9 +1,35 @@
...
+  HW.shout('rawr', {}, function(err, shout) {
+    if (err) throw err;
+    console.log(shout);
+    t.end();
+  });
+});
+
+test('shout error - non string phrase', function(t) {
+  HW.shout(4, {}, function(err, shout) {
+    t.ok(err);
```
mapsam: I would expect `err` to be here, but we are throwing within `hello_world.cpp` instead of returning an error message. Not sure how to do this. @springmeyer

mapsam: Actually on second thought, maybe throwing is best here and the test should be written differently. Not sure.

springmeyer: Right, the current code potentially both throws and returns an error in the callback. So the test is incorrect and should do a `try/catch` instead and assert the catch is hit.

This "two ways to catch errors" is not ideal and results in extra code in JS of course. But it is not viable to do the validation in the `HelloWorld::AsyncShout` because one cannot safely copy the data into the threadpool without validating.

Potentially one way to avoid going into the threadpool/still return all errors in the callback would be to call the callback inside `HelloWorld::shout` instead of throwing errors. I think this should work but I've not tried it. Would be a great thing to look into for this repo because figuring out a cleaner way to do this would benefit the usability of all modules based on this code. Can you ticket for a little later?

src/hello_world.cpp
```diff
@@ -144,27 +145,25 @@ NAN_METHOD(HelloWorld::shout)
...
-        return_string += "!!!!";
-    }
 
-    /***************** end custom code *******************/
+        std::string return_string = baton->phrase + "!";
 
-    try
-    {
-        baton->result = return_string;
+        if (baton->louder)
```
mapsam: @springmeyer seeing that the `return_string` now [comes back empty](https://travis-ci.org/mapbox/node-cpp-skel/jobs/138593983#L338) for this method after wrapping it all within the `try/catch`. Seems odd, but if I pull it out:

```
/***************** custom code here ******************/

std::string return_string = baton->phrase + "!";

if (baton->louder)
{
    return_string += "!!!!";
}

/***************** end custom code *******************/

try
{
    baton->result = return_string;
}
catch (std::exception const& ex)
{
    baton->error_name = ex.what();
}
```

things work fine

src/hello_world.cpp
```diff
@@ -26,17 +32,161 @@ NAN_METHOD(HelloWorld::New)
...
+    /***************** custom code here ******************/
+    try
+    {
+
+        std::string return_string = baton->phrase + "!";
+
+        if (baton->louder)
+        {
+            return_string += "!!!!";
+        }
```
mapsam: ah, I see what the issue was @springmeyer - we lost the `baton->result = return_string` here. Adding it back in does the trick.

mapsam:
All green @springmeyer - this is ready to be merged.

springmeyer:
Great, looks good to me to merge as well - we can address future improvements as they arise. Notably integrating the results of https://github.com/Project-OSRM/node-osrm/pull/193#issuecomment-227101937, when I understand more.

PR #12, opened 2016-06-23
More convenient builds

GretaCB:
Per chat with @mapsam and @springmeyer , soup up the skeleton's Makefile with faster builds and "gotchyas" prevention.

Features:
- revised `default`: When running `make`, speeds up rebuilds when all that has changed is the cpp file you are editing. Skips rerunning npm install which is slow. This makes iterating through development faster as you make changes and test your changes.
- `make clean`: removes npm and mason deps
- `$(MODULE_NAME)`: convenience configuration/generalization

cc @mapsam @springmeyer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,17 @@
-default:
-	npm install --build-from-source
+MODULE_NAME := $(shell node -e "console.log(require('./package.json').name)") 
+
+default: node_modules
+	./node_modules/.bin/node-pre-gyp build
 
 debug:
-	npm install --build-from-source --verbose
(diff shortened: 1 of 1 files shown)


mapsam:
Woo woo looks great @GretaCB!

springmeyer:
@GretaCB looks great. You can merge! Just one more suggestion: pass `--loglevel=error` by default to `node-pre-gyp build` (note it could also be added to `npm install` but since that will only be run once, its less valuable there). This will make all the npm logging much less verbose such that when there is a C++ compile error (as you'd frequently expect during development) you will have to scroll through much less npm noise to see it.

PR #13, opened 2016-06-29
Add Xcode project

kkaefer:
We should provide a way to create an Xcode project that compiles the node module and adds a schema to run `npm test`. This makes debugging node modules way easier on macOS.

/cc @mapsam @springmeyer @GretaCB


mapsam:
Thanks @kkaefer!

Issue #14, opened 2016-07-13
Pre-built binaries on a remote


mapsam:
Is it worth documenting how to deploy pre-built binaries for each system architecture? Right now the [`package.json` points to a remote location](https://github.com/mapbox/node-cpp-skel/blob/master/package.json#L24-L25) but as far as I can tell there's nothing to build and deploy to this location.

cc @springmeyer @jakepruitt @GretaCB

mapsam:
Since this will come up for any cpp project, I think it's probably worth adding a commit-based script for publishing binaries via travis, something similar to Node Mapnik: https://github.com/mapnik/node-mapnik/blob/master/scripts/publish.sh

mapsam:
Sooooo the issue here is that pre-built binaries on an s3 bucket will require a specific set of secret tokens, etc. Those can't be shared across projects (in the sense that we are sharing code with node-cpp-skel), but we _can_ document how to set it up. The access to `https://mapbox-node-binary.s3.amazonaws.com` is Mapbox user specific, so maybe we should make a note in the README about pre-built binary releasing that discusses what a user needs to do to get things set up.

mapsam:
Okay, I've been working on getting this set up with `vtinfo` and think we can point to the node-pre-gyp docs about setting up secure access keys for travis: https://github.com/mapbox/node-pre-gyp#travis-automation

Probably best to link to the "how to generate the proper IAM roles" as well, so people can learn more about that.

springmeyer:
> think we can point to the node-pre-gyp docs about setting up secure access keys for travis: https://github.com/mapbox/node-pre-gyp#travis-automation

Correct.

> Probably best to link to the "how to generate the proper IAM roles" as well, so people can learn more about that.

👍  What is critical is that the role is scoped to the specific bucket + module such that `npm unpublish` or `npm publish` cannot overwrite any binaries inside the bucket not related to the specific module.

mapsam:
Going to confirm publishing binaries works as expected, then we can update the readme, then close this!

mapsam:
Also look at this commit - confirm it works: https://github.com/mapbox/node-cpp-skel/commit/dcd6e49faff2a851164c1357a7af2cc6f3ddea78

mapsam:
Publishing now works for linux. Going to wait for the OSX build to finish 🍏  before merging.

mapsam:
OSX is passing! Now waiting for binaries to publish for it ... then we're off to the races. 🏇

[...]

PR #15, opened 2016-07-13
include publish binary scripts

mapsam:
Working on #14 

This adds a bit more cruft, but seems pretty worthwhile if it allows others to actually `npm install` and use the library?

cc @springmeyer @mikemorris @GretaCB @jakepruitt


mikemorris:
I rigged up [a script to only publish on git tags](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/node/scripts/after_script.sh) for mapbox-gl-native, but unfortunately it doesn't support Bitrise, as it relies on the [`TRAVIS_TAG`](https://github.com/mapbox/mapbox-gl-native/blob/master/.travis.yml#L68) environment variable. Figuring out some convention that works across Travis CI, Bitrise and Circle CI here would be great.

springmeyer:
👍  Great stuff @mapsam.

Also, I think we should start refactoring all logic into a single bash file. Then the convention can be to source this single bash file for its functions and call them. So, what do you think about combining `scripts/is_pr_merge.sh` and `scripts/publish.sh` into a single `scripts/common.sh` that could be easily copied from this repo to other node c++ projects? My inspiration for this is https://github.com/mapnik/mapnik/blob/master/scripts/travis-common.sh

> Figuring out some convention that works across Travis CI, Bitrise and Circle CI here would be great.

👍

mapsam:
Oooh @mikemorris I like the idea of building ONLY on tags a lot. I've included the [logic in `vtinfo` to NOT build on tags](https://github.com/mapbox/vtinfo/pull/4/files#diff-354f30a63fb0907d4ad57269548329e3R5). Could we do a combination of build on tags AND an option to republish binaries if necessary?

mikemorris:
@mapsam Yea-ish. That's what how [the binary publishing script in node-fontnik](https://github.com/mapbox/node-fontnik/blob/master/scripts/travis_publish.sh#L16) is currently written, but the problem is that the equivalent `BITRISE_GIT_TAG` [doesn't work as expected](https://github.com/mapbox/mapbox-gl-native/issues/4854). Would be great to get this figured out and fixed!

mapsam:
Merging this for now - going to open separate tickets about further consolidation and circleci/bitrise additions.

springmeyer:
👍  @mapsam

Issue #25, opened 2016-10-10
Create benchmark scaffolding


GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/17, this also enables profiling/performance tracking to work off from.

cc @springmeyer @mapsam

springmeyer:
Solved in #61

PR #29, opened 2016-10-11
[WIP] Add benchmark scaffolding

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/25
- [x] Add single run test
- [x] Add batch run test
- [x] Try out `process.memoryUsage()`. Thinking about [what info is useful here](http://stackoverflow.com/questions/12023359/what-do-the-return-values-of-node-js-process-memoryusage-stand-for)
- [x] Add `sleep` option for mocking when threads are busy, but aren't doing much work
- [x] Add more heavy-duty benchmarks per https://github.com/mapbox/node-cpp-skel/issues/30#issuecomment-253306888

Per chat with @springmeyer:
- Can we write a test that proves memory optimization?
- node module-ify @springmeyer 's [libnew](https://github.com/springmeyer/libnew) lib for tracking of memory allocations
- [Document how to profile](https://github.com/mapbox/node-cpp-skel/issues/30) (using Activity Monitor). The skel isn't doing heavy operations, but might be possible to profile during batch bench run.

cc @mapsam @springmeyer


test/bench/bench-batch.js
```diff
@@ -0,0 +1,59 @@
+var fs = require('fs');
+var path = require('path');
+var argv = require('minimist')(process.argv.slice(2));
+var tape = require('tape');
```
GretaCB: 👍  awesome thank you! just committed the change.

springmeyer:
@GretaCB https://github.com/mapbox/node-cpp-skel/pull/29/commits/cf913b8adde27d14531644ce4d9b8961b408ffc9 adds a very expensive usage of `std::map` (because it invokes lots of memory allocation internally in the map, searching of the map, and string comparisons). Now with:

```bash
~/projects/node-cpp-skel[bench]$ time node test/bench/bench-batch.js  --iterations 10 --concurrency 10

real	0m3.491s
user	0m11.704s
sys	0m0.535s
```

I get my CPU usage spiking to > 500%. Running `node test/bench/bench-batch.js  --iterations 100 --concurrency 10` to keep it going long enough to easily attach in activity monitor gives a callstack 98% idle in the main event loop (as expected if the threads are doing all the work) and with 99.9-100% of the threads reporting busy doing work (:tada:):

screen shot 2016-10-18 at 4 02 53 pm

springmeyer:
per chat with @GretaCB - next I'm going to take a look at profiling and tuning a few things in the PR. In particular I'll look at the impl of the mutex lock and make sure there is enough work being done in the async function that locks the global such that we are properly demonstrating (e.g. providing an example you can profile) the kind of thread contention programmers should avoid.

springmeyer:
> In particular I'll look at the impl of the mutex lock and make sure there is enough work being done in the async function that locks the global such that we are properly demonstrating (e.g. providing an example you can profile) the kind of thread contention programmers should avoid.

Done in https://github.com/mapbox/node-cpp-skel/commit/fb1fca8ee7ddfe6660fb6147dfa8f1bb955500a5. Now the `contentiousThreads` demo is properly awful. It can be tested like:

```
node test/bench/bench-batch.js --iterations 50 --concurrency 10 --mode contentiousThreads
Benchmark speed: 15 runs/s (runs:50 ms:3245 )
Benchmark iterations: 50 concurrency: 10 mode: contentiousThreads
```

If you bump up `--iterations` to 500 and profile in Activity Monitor.app you'll see the main loop is idle. This is expected because it is only dispatching work to the threads. The threads however are all busy in `psynch_mutexwait` (waiting for a locked mutex) as more time is spent waiting than doing the expensive work. This is because one thread will grab a lock, do work, all others will wait, another will grab the released lock, do work, all other threads will wait. This is all too common and the reason you don't want to use `mutex` locks. This is the profiling output of this non-ideal situation:

screen shot 2016-11-03 at 5 43 27 pm

When locks are unavoidable in real-world applications, we would hope that the % of time spent in `psynch_mutexwait` would be very small rather than very big. The real-world optimization would be to either rewrite the code to avoid needing locks or at least to rewrite the code to hold onto a lock for less time (scope the lock more).

springmeyer:
Just looked back at this branch. It's got some great stuff that I think we should merge soon and keep iterating on. My one reservation before merging I'm slightly uncomfortable with how we are mixing best practice/simple/hello-world style code with new code demonstrating both advanced and non-ideal scenarios. I think we should split things apart before merging such that:

  - `src/hello_world.hpp` and `src/hello_world.cpp` contain only simple hello world style code: only binding a single standalone function (this is what you would copy if you just want to start writing a module fast and use the skel as a base)
  - `src/async_<name>.hpp` and ``src/async_<name>.cpp` and `src/async_<name>.readme.md` that demonstrate common async scenarios in code and have a readme talking about what about them to care about them (good and bad). (This is what you'd profile, and poke at within skel, to learn about the nuances of node async performance).

springmeyer:
@GretaCB I just returned here to reflect on next steps. I feel like a good approach would be to split this work into 2 phases:

### phase 1

Land a first PR that:

 - [ ] Adds a `./bench` folder
 - [ ] Adds two benchmark scripts to cover the two performance focused functions currently in master:
   - [ ] `./bench/hello_async.bench.js`
   - [ ] `./bench/hello_object_async.test.js`
 - [ ] These scripts could be based on a simplified version of the `test/bench/bench-batch.js` currently in this PR
 - [ ]  Very simple docs added to the README on how to run the benchmark and a comment that it is not realworld yet since the code does not do much, but that the idea is for developers using skel to adapt it and run it to monitor the performance of the code they add.

^^ this gets the key structure in place to make it easy and fast to start benchmarks for any module that uses skel. This is a great first step.

### phase 2

We could revisit adding examples of performance scenarios. However I feel like this is a really advanced topic most suitable outside of node-cpp-skel. The skel is complex enough currently without diving deep on performance. But given performance is critical it would be great to cover it and benefit from the skel structure. So, here is an idea. Instead of building into the skel directly, we could:

 - Fork skel, and remove lots of unneeded code and docs (like the sync examples)
 - Add back key async performance scenarios
 - Add docs for this
 - Then link to this fork as an example of a skel implementation
 - The fork would be a separate "performance deep dive example using node-cpp-skel"

springmeyer:
@GretaCB  now that Phase 1 is done in #61 - how about closing this ticket? Phase 2 remains but I feel like my idea is not concrete enough to warrant a ticket. I'm feeling really good with what we have and don't see a major need to ticket more work here. Rather we'll apply node-cpp-skel, learn perf issues, and then - at that time - have ideas of things to build back, or add to the docs.

springmeyer:
Closing this. What we have are:

 - Benchmark scripts: https://github.com/mapbox/node-cpp-skel/tree/master/bench
 - Benchmarking docs: https://github.com/mapbox/node-cpp-skel/blob/master/docs/benchmarking.md
 - Future performance questions to answer like #143

PR #33, opened 2016-10-13
remove magic NAN macros

### LINKED ISSUE #17: Create a branch without magic NAN macros, as recommended in newer docs
Per chat with @mikemorris @springmeyer @GretaCB - there are some new approaches to working with NAN that remove the usage of magic macros like `NAN_METHOD`. Let's create a branch with [information from this blog post](https://nodesource.com/blog/c-add-ons-for-nodejs-v4/) and see what it looks like!

mikemorris:
Fixes #17

And adds a few modern NAN best practices from https://nodesource.com/blog/cpp-addons-for-nodejs-v4/

/cc @mapsam @gretacb @springmeyer

### DIFF
--- a/src/hello_world.cpp
+++ b/src/hello_world.cpp
@@ -20,6 +20,8 @@ HelloWorld::HelloWorld(std::string name) :
  *          OR
  * var HW = new HelloWorld('yo');
  */
+
+// NAN_METHOD(HelloWorld::New)
 void HelloWorld::New(const Nan::FunctionCallbackInfo<v8::Value>& info)
 {
     if (info.IsConstructCall())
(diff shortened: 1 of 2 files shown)


mikemorris:
It looks like this doesn't run on Node.js 0.10 (`master` doesn't seem to either), although it probably could. Is this something worth fixing?

```
module.js:356
Module._extensions[extension](this, filename);
                               ^
Error: Symbol hello_world_module not found.
```

mapsam:
Nice @mikemorris! Thanks for putting this together. For the sake of the skel, I think it would be useful to keep the macros as comments above, just as a reference for later. What do you think?

springmeyer:
I've reviewed this and the https://nodesource.com/blog/cpp-addons-for-nodejs-v4/. I feel like the `NAN_METHOD` macros are fine: they are still supported, more concise, and work fine. So I'm not feeling like this improves the code significantly. However the use of `Nan::Utf8String` is smart and worthwhile to land. Much refactoring has happened recently, so this PR is stale. I'm going to close this now and have created a ticket for tracking moving to `Nan::Utf8String` at #54

PR #41, opened 2017-03-16
Document how to set up credentials for publishing

springmeyer:
### Context

node-cpp-skel offers a working example for how to create a node c++ module, including scaffolding for critical pieces like publishing the binaries with node-pre-gyp. So, node-cpp-skel is the home to docs on best practices and also a working example.

### Need

We need to document how to setup credentials correctly for publishing. The assumption is that the docs will be followed primarily by Mapbox staff, but the could also be used by anyone to publish to any bucket.

TODO:

  - [x] Doc review: would love 👀  from @alulsh and @mapsam
  - [x] Make `node-cpp-skel` scoped to `@mapbox`. Ensure publishing still works as expected.
    - Should publishing be to `s3://<bucket>/@mapbox/<module name>` or `s3://<bucket>/<module name>`? Currently the docs will break with publishing to `@mapbox` so that needs fixed.
  - [ ] Work to make the methods more automated, easier to replicate in other repos (this will be applied to dozens).


README.md
```diff
@@ -60,6 +60,61 @@ It's a good idea to publish pre-built binaries of your module if you want others
...
+npm install @mapbox/cfn-config
+npm install @mapbox/cloudfriend
+./node_modules/.bin/validate-template cloudformation/ci.template.js
+./node_modules/.bin/build-template cloudformation/ci.template.js > cloudformation/ci.template
+./node_modules/.bin/cfn-config create ci-binary-publish cloudformation/ci.template -c cfn-configs
+```
+
+3) Encode aws keys in .travis.yml
+
+First export your config with mbxcli (`mbx export-stack-config ci-binary-publish`).
```
TheMarex: This fails for non-mapbox org repositories. (e.g. `project-osrm`) with:
> Could not determine service name from git@github.com:Project-OSRM/osrm-backend.git. Is it under the mapbox organization on Github?

cloudformation/ci.template.js
```diff
@@ -0,0 +1,54 @@
...
+      Properties: {
+        Policies: [
+          {
+            PolicyName: 'list',
+            PolicyDocument: {
+              Statement: [
+                {
+                  Action: ['s3:ListBucket'],
+                  Effect: 'Allow',
+                  Resource: 'arn:aws:s3:::mapbox-node-binary/' + package_json.name
```
TheMarex: This policy is wrong and will not work because the resource name is wrong. Should be:

```
          {
            PolicyName: 'list',
            PolicyDocument: {
              Statement: [
                {
                  Action: ['s3:ListBucket'],
                  Effect: 'Allow',
                  Resource: 'arn:aws:s3:::mapbox-node-binary',
                  Condition : {
                    StringLike : {
                      "s3:prefix": [ package_json.name + "/*"]
                    }
                  }
                }
              ]
            }
          },

```

Without this publishing will fail with a weird `Forbidden` error, because `node-pre-gyp` calls `HeadObject` on a (potentially) non-existent file. Only if you have `ListBucket` rights this will return `NotFound` instead of `Forbidden`.

springmeyer: 👍  thank you @TheMarex. Will integrate now.

daniel-j-h:
What's the progress here @springmeyer?

I was looking into this today and only found this as a guide on how to generate tokens.

Also there should maybe be a section on how to get the tokens in the node-pre-gyp readme, here: https://github.com/mapbox/node-pre-gyp#travis-automation
It describes how to encrypt keys but not how to get the keys in the first place.

cloudformation/ci.template.js
springmeyer: Fixed in 06b1087

springmeyer:
Update on this:

 - I've revised the publishing section in the readme to be more detailed and less `mbxcli` specific.
 - I've removed the keys from the `travis.yml` since the docs now recommend putting these in the travis settings UI.
 - I've pushed a commit to put the module in the `@mapbox` namespace

The next steps here are to:

  - [ ] @mapsam runs through my latest docs once more, make sure they look good
  - [ ] @mapsam actually runs through using the docs to create a user for this repo and sets up publishing. (Currently publishing will be broken after I removed the keys in this branch)
  - [ ] Once publishing is confirmed to be working again in this branch, @mapsam merges this PR.

README.md
springmeyer: I've revised the docs to show how to do this with the AWS UI instead. Hopefully that will resolve this issue: https://github.com/mapbox/node-cpp-skel/pull/41/commits/ffe3deeb8f4227e40ad9a571a0a80f58b6535165. @TheMarex can you test again?

README.md
```diff
@@ -55,10 +55,120 @@ console.log(hi); // => howdy world!
...
+6) Add the keys to the travis
+
+You can do this two ways: 1) add the keys to the travis UI settings, or 2) encode them as secure variables in your `.travis.yml`
+
+#### Adding to travis UI settings
+
+- Go to https://travis-ci.org/<your user or org>/<your module>/settings
+- Scroll to the bottom and find the `Environment Variables` section
+- Add a variable called `AWS_ACCESS_KEY_ID` and put the value of the `AccessKeyId` in it
+- CRITICAL: Choose `OFF` for `Display value in build log` to ensure the variables are now shown in the logs
```
mapsam: We don't want to show our access keys in the logs on travis, right? I think we should keep critical to `ON` here. The logs will still show which variables exist, but if critical is turned `OFF` it will show the variable AND the value of that variable.

README.md
```diff
@@ -55,10 +55,120 @@ console.log(hi); // => howdy world!
...
+```
+
+4) Create a user with permissions to upload to `s3://mapbox-node-binary/<your module name>/`
+
+First configure your AWS creds. You will need to set at least the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` keys. And also `AWS_SESSION_TOKEN` if you are using 2-factor auth.
+
+Then run:
+
+```bash
+./node_modules/.bin/validate-template cloudformation/ci.template.js
```
mapsam: Since `@mapbox/cloudfriend` isn't a dependency of this module this fails since the ci.template.js requires `require('@mapbox/cloudfriend')`. Happy to:

A. include the dependency, or
B. write the template without cloudfriend

What do you think @springmeyer?

mapsam: With option B, we can just remove this validation step entirely and have the raw JSON as the template to use.

README.md
```diff
@@ -55,10 +55,120 @@ console.log(hi); // => howdy world!
...
+```
+
+It will prompt you, choose:
+
+  - New configuration
+  - Ready to create the stack? Y
+
+It will fail if the stack already exists. In this case you can recreate a new user by deleting the stack by running `./node_modules/.bin/cfn-config delete ci-binary-publish cloudformation/ci.template` and then creating a new one.
+
+5) Get the user keys
```
mapsam: You can run `cfn-config info <stack_name>` to get this information as well. In this case, it looks like:

```
~/mapbox/node-cpp-skel[cf]$ cfn-config info ci-binary-publish
{
  "StackId": "arn:aws:cloudformation:us-east-1:......",
  "StackName": "node-cpp-skel-ci-binary-publish",
  "Description": "user for publishing to s3://mapbox-node-binary/@mapbox/node-cpp-skel",
  "Parameters": {},
  "CreationTime": "2017-06-22T20:48:35.469Z",
  "StackStatus": "CREATE_COMPLETE",
  "DisableRollback": false,
  "NotificationARNs": [],
  "Capabilities": [
    "CAPABILITY_IAM",
    "CAPABILITY_NAMED_IAM"
  ],
  "Outputs": {
    "AccessKeyId": "...",
    "SecretAccessKey": "..."
  },
  "Tags": {},
  "Region": "us-east-1"
}
```

README.md
mapsam: Oh ... nvm I was interpreting `OFF` and `ON` in the reverse direction. Maybe we should just make a note saying "Do not display the token in the logs!"

README.md
springmeyer: > write the template without cloudfriend

The benefit of `cloudfriend` is the ability to dynamically grab the `package.json`'s `name` property. This is critical so that the permissions are scoped to the specific module and not the root directory. Fewer deps is better, but not sure how to do this without cloudfriend. Can you think of a way to do this without `cloudfriend`?

> include the dependency

My reasoning in leaving this off (and only mentioning to install them in the `step 3` above) was that this is a one-time thing. But if you think it would be better to add to devDeps, I'm 👍

README.md
springmeyer: oooo, I did not realize that. Could you add that?

README.md
springmeyer: Sorry for the confusion. Yes, it should be `OFF` but I also see there is a typo. It should be:

```
- to ensure the variables are now shown in the logs
+ to ensure the variables are NOT shown in the logs
```

> Maybe we should just make a note saying "Do not display the token in the logs!"

👍 Please feel free to freely edit the text to make this clearer. 🙏

mapsam: Ah, that makes a lot more sense now 😄 

> feel free to freely edit the text

will do!

README.md
mapsam: Yeah, cloudfriend is super nice to have and using the package.json is a great idea. Right now the `./node_modules/.bin/validate-template` command just doesn't work unless the dependency is added. Even when I have cloudfriend installed globally requiring doesn't work 🤔

springmeyer: How about recommending a global install and then removing the `./node_modules/.bin/` so that it actually works? The `npm install -g @mapbox/cloudfriend` could be moved to `Step 4` to reduce the chance it is missed.

mapsam: I tried this, and have cloudfriend installed globally but am still getting a "module not found" error when running validate-template

springmeyer: I tried the below and it worked for me:

```
$ npm install -g @mapbox/cloudfriend
/Users/dane/.nvm/versions/node/v4.8.3/bin/validate-template -> /Users/dane/.nvm/versions/node/v4.8.3/lib/node_modules/@mapbox/cloudfriend/bin/validate-template.js
/Users/dane/.nvm/versions/node/v4.8.3/bin/build-template -> /Users/dane/.nvm/versions/node/v4.8.3/lib/node_modules/@mapbox/cloudfriend/bin/build-template.js
@mapbox/cloudfriend@1.8.1 /Users/dane/.nvm/versions/node/v4.8.3/lib/node_modules/@mapbox/cloudfriend
└── aws-sdk@2.77.0 (sax@1.2.1, xml2js@0.4.17, querystring@0.2.0, uuid@3.0.1, jmespath@0.15.0, url@0.10.3, buffer@5.0.6, crypto-browserify@1.0.9, xmlbuilder@4.2.1)
$ which validate-template
/Users/dane/.nvm/versions/node/v4.8.3/bin/validate-template
$ validate-template cloudformation/ci.template.js 
✘ invalid: Missing credentials in config # would have worked if I had auth'd
```

mapsam: Weird, here's my output: 

```
$ which validate-template
/Users/mapsam/.nvm/versions/node/v4.4.7/bin/validate-template
$ validate-template cloudformation/ci.template.js
module.js:327
    throw err;
    ^

Error: Cannot find module '@mapbox/cloudfriend'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
    at require (internal/module.js:12:17)
    at Object.<anonymous> (/Users/mapsam/mapbox/node-cpp-skel/cloudformation/ci.template.js:1:72)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
    at Module.require (module.js:353:17)
~/mapbox/node-cpp-skel[cf]$ 
```

README.md
mapsam: Updated!

README.md
mapsam: Added!

mapsam:
@springmeyer I made some edits and have synced with master. Should we attempt to publish a binary from this branch or wait until we're in master?

springmeyer:
> @springmeyer I made some edits and have synced with master.

🙌 

> Should we attempt to publish a binary from this branch or wait until we're in master?

We should test that publishing works. Your call whether this is done before/after merge. Please run this out as you see fit.

One note is that I see a few travis failures that are mysterious:

```
not ok 2 test exited without ending
  ---
    operator: fail
```

I think these will have been resolved in master by removing `node-pre-gyp` from being used dynamically in the `require` after https://github.com/mapbox/node-cpp-skel/pull/44/files#diff-474b37290c5ca9ae5a02270ddfd00d96. So, I would recommend rebasing this branch against `master` and making sure travis passes after: `git rebase -i master`. Hopefully it should. If not, it is still not likely a problem with this branch and we can continue with a merge. Just would be nice to see things green.

mapsam:
Hey @springmeyer - I synced with master yesterday and rerunning those two tests made them green, not sure why they failed. Looking good now!

> We should test that publishing works.

Let's do this once we merge into master, as if we are doing a normal release.

springmeyer:
> Let's do this once we merge into master, as if we are doing a normal release.

👍 Okay, then please merge this now if you've landed all your edits. 🎉 

I figure there might be glitches when this approach is applied to other repos (say the node-mapnik repo in a different org) but we can make fixes when/if we hit those.

Issue #10, opened 2016-06-22
expose an example standalone function


mapsam:
We don't always need to functions to exist as full object constructors (i.e. `var waka = new Flocka()`. Let's put together an example of a module that is just a standalone function.

GretaCB:
Per chat:
- In vt-shaver, we implemented a standalone function, which is more performant because you dont need to create individual object instances for each use. It's a micro-optimization since object instances require an allocation and garbage collection to track them. We're preventing V8 from having to do unnecessary work.
- Build a standalone hello-world function as "hello-world.cpp" as default, and move current code into hello_world_class.cpp

mapsam:
Looking at nodejs's example repo, they have it structured for different types of javascript functionality: https://github.com/nodejs/node-addon-examples

GretaCB:
All set per https://github.com/mapbox/node-cpp-skel/pull/44

PR #44, opened 2017-06-14
Standalone function refactor

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/10 and due to the ever-expanding uses-cases and complexity within [the bench branch](https://github.com/mapbox/node-cpp-skel/pull/29), there's a clear need to simplify the skel. 

This PR is the start of refactoring the skel to allow for less interconnected code, making it easier to scale and to add new and separate implementations. Each new level of complexity can have its own `.cpp` and `.hpp` files within `/src/your_code_here/`, and pull them all together via `module.cpp`

That way, if you fork node-cpp_skel, we can document what you'd need to do:
- edit binding.gyp source array
- edit SetMethod
- delete any unneeded cpp/hpp files 

### Next
- [x] Decide what to do with class-based `hello-world.cpp`
  - Name it `class-example`?
  - Slim it down to a single async function?
  - Is there a need for a class example?
- [x] Document how to customize when adding new standalone functions
- [x] Incorporate NAN best practices per https://github.com/nodejs/nan/blob/master/doc/node_misc.md#api_nan_module_init?
  - we're all set for now, will do a future pass for NAN where necessary
- [x] Incorporate async-specific updates per https://github.com/nodejs/nan/blob/master/doc/asyncworker.md#api_nan_async_queue_worker? 
  - Will implement this in the upcoming `async` examples

cc @springmeyer @mapsam


lib/hello_world.js
```diff
@@ -1,9 +0,0 @@
-"use strict";
```
GretaCB: @springmeyer What was the purpose of lines 3-6?

src/hello.cpp
```diff
@@ -0,0 +1,10 @@
+#include "hello.hpp"
+
+// Standalone function (because it's not a class)
+NAN_METHOD(hello) {
+
+    // info comes from the NAN_METHOD macro, which returns differently
+    // according to the version of node
+    info.GetReturnValue().Set(Nan::New<v8::String>("world").ToLocalChecked());
+
+}
```
GretaCB: @springmeyer possible to add another Method in here? Or is the concept of "standalone function" that there is only _one_ function as part of this `cpp` file?

test/hello_world.test.js
```diff
@@ -1,120 +0,0 @@
-var test = require('tape');
```
GretaCB: @springmeyer Shall I plan to re-add this to a separate branch, like `class-example`?

src/standalone/hello.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <nan.h>
+// carrots, ex: <nan.h> --> look for header in global
+// quotes, ex: "hello.hpp" --> look for header in location relative to this file
```
mapsam: so helpful!

binding.gyp
```diff
@@ -28,7 +28,7 @@
       'target_name': '<(module_name)',
       'product_dir': '<(module_path)',
       'dependencies': [ 'action_before_build' ],
-      'sources': [ './src/hello_world.cpp' ],
+      'sources': [ './src/module.cpp', './src/hello.cpp' ],
```
springmeyer: We could also do a "glob" here. Something like: 

```
"<!@(find ./src -name '*.cpp')"
```

That would avoid needing to add files explicitly. If you think this makes sense @GretaCB - I'd recommend adding this.

Caveat however: is that `find` won't work on windows. I don't think we need to support windows, but if we ever did we could use the cross-platform approach using python at https://github.com/mapnik/mapnik-gyp/blob/d8b21e7698334fc9e574d330ed8bd9db061cb8e1/mapnik.gyp#L122

lib/hello_world.js
springmeyer: To be able to find the binary when the `module_path` has versioning variables. Because we just use `lib/binding` [here](https://github.com/mapbox/node-cpp-skel/blob/59d00921ce3751b7a688ee22c250e99c74e9b34a/package.json#L30) and not versioning variables like: https://github.com/mapbox/node-sqlite3/blob/f1456b1cc96e89b82ac37b1896f23ef14aff80b6/package.json#L12 we can safely hardcode this path. Overall the ability to use versioning variables in node-pre-gyp I consider an edge case feature/too much flexibility and should likely be removed from node-pre-gyp in future versions. The motivation/need for versioning varables in the `module_path` is to be able to install multiple binaries for multiple node platforms and have them both running locally. Then you could `nvm use <diff version>` and be able to run your app without re-installing. But again, this is a rare need.

src/hello.cpp
springmeyer: We could absolutely add another function here in the same cpp file. The concept of `standalone function` only relates to "let's have a function that is not a method on a class, but rather on its own". In this case the `standalone function` would be in the global scope. If namespaces were used it would be in the namespace (http://en.cppreference.com/w/cpp/language/namespace).

In fact I think it would be a good idea to drop this in a namespace from the start. That would look like:

 - Adding `namespace <some name> { ... }`
 - Putting the function inside the namespace
 - Changing the `Nan::SetMethod(target, "hello", hello);` in `src/module.cpp` to `Nan::SetMethod(target, "hello", somename::hello);`

springmeyer: Oh, and also wrapping the `NAN_METHOD(hello);` in the `hello.hpp` in a namespace.

src/hello.hpp
```diff
@@ -0,0 +1,8 @@
+#pragma once
+#include <nan.h>
+// carrots, ex: <nan.h> --> look for header in global
+// quotes, ex: "hello.hpp" --> look for header in location relative to this file
```
springmeyer: Good explanation for this at https://stackoverflow.com/a/21594.  @GretaCB - it would be great if you could add a mention to this in the C++ glossary (https://github.com/mapbox/cpp/blob/master/glossary.md).

test/hello_world.test.js
springmeyer: Yes, would be great to add a class example back. But, I would put adding an example of an `async standalone` function higher priority in your list than a class example (partly because the latter is more complex and less frequently a need).

src/standalone/hello.hpp
springmeyer: See also https://stackoverflow.com/a/21594

binding.gyp
GretaCB: Lets stay explicit, but I'll add a comment link to this suggestion 👍

test/hello_world.test.js
GretaCB: Per chat with Dane, this will be a separate subdirectory. I plan to implement this (as well as more example) in future branches that will be merged into master...in due time.

GretaCB:
Hey @springmeyer , just FYI, I [integrated the HelloWorld class](https://github.com/mapbox/node-cpp-skel/pull/44/commits/d10e3c56b30b12995bede9ca8a3e09dadc5e28fc) and re-added the original tests to retain them in master so that this merge is more of an addition than an entire sweep. 

I noticed [some funky behaviour in Travis](https://travis-ci.org/mapbox/node-cpp-skel/jobs/243065289#L373) that I'm curious about:
- Tests worked fine locally after a fresh distclean, and running `console.log(module)` returned:
```
{ hello: [Function: hello],
  HelloWorld: [Function: HelloWorld],
  version: '1.0.0' }
```
- Tests failed in Travis due to the `hello` function not being recognized, and running `console.log(module)` returned:
```
{ HelloWorld: [Function: HelloWorld], 
  version: '1.0.0' }
```

**For some reason, the `hello` function is never added to the module.** I removed `./src/hello_world.cpp` from the build, so everything is working fine now, but curious why Travis was different.

GretaCB:
Above ^^^ was due to being on node `v0.10.x` instead of node `v4`.

PR #38, opened 2017-02-16
Use Async Workers for Heavy Async. Tasks

daniel-j-h:
This repository is great for getting people up to speed with Node bindings and Nan.

Let's switch over to Nan's AsncWorkers for async. tasks.
It's simpler, clearer and less error prone then doing it ourselves with uv.

Somewhat related: #5.

cc @springmeyer

PR #46, opened 2017-06-15
Add async standalone function example

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/pull/44 and https://github.com/mapbox/node-cpp-skel/pull/38, this PR adds an _async_ standalone function example.

Note: It was super easy to integrate a new example into the current structure 🙌 

### Next
- [x] Add docs for this example
- [x] Add inline comments to give more context for using NAN's AsyncWorker

@daniel-j-h Mind taking a peek?

cc @springmeyer @mapsam

### DIFF
--- a/src/standalone_async/hello_async.cpp
+++ b/src/standalone_async/hello_async.cpp
@@ -35,6 +35,9 @@ namespace standalone_async {
       for (std::size_t i=0;i<work_to_do;++i) {
           std::string const& item = container[i];
           if (item != std::to_string(i)) {
+
+              // AsyncHelloWorker's Execute function will take care of this error 
+              // and return it to js-world via callback
               throw std::runtime_error("Uh oh, this should never happen");
           }
       }  
--- a/src/standalone_async/hello_async.cpp
+++ b/src/standalone_async/hello_async.cpp
@@ -60,7 +60,7 @@ namespace standalone_async {
       using Base = Nan::AsyncWorker;  
 
       AsyncHelloWorker(bool louder, Nan::Callback* callback)
-          : Base(callback), louder_(louder) { }  
+          : Base(callback), result_{""}, louder_{louder} { }  
 
       // The Execute() function is getting called when the worker starts to run.
       // - You only have access to member variables stored in this worker.
--- a/src/standalone_async/hello_async.cpp
+++ b/src/standalone_async/hello_async.cpp
@@ -1,0 +1,137 @@
+#include "hello_async.hpp"
+
+#include <exception>
+#include <stdexcept>
+#include <iostream>
+#include <map>
+
+namespace standalone_async {
+
+  /*
+  * This is an internal function used to return callback error messages instead of
+  * throwing errors.
+  * Usage:
+  *
+  * v8::Local<v8::Function> callback;
+  * CallbackError("error message", callback);
+  * return; // this is important to prevent duplicate callbacks from being fired!
+  */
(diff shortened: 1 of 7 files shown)


src/standalone_async/hello_async.cpp
```diff
@@ -0,0 +1,137 @@
...
+namespace standalone_async {
+
+  /*
+  * This is an internal function used to return callback error messages instead of
+  * throwing errors.
+  * Usage:
+  *
+  * v8::Local<v8::Function> callback;
+  * CallbackError("error message", callback);
+  * return; // this is important to prevent duplicate callbacks from being fired!
```
daniel-j-h: `return CallbackError("error message", callback)` works, too

src/standalone_async/hello_async.cpp
```diff
@@ -0,0 +1,137 @@
...
+      std::map<std::size_t,std::string> container;  
+
+      for (std::size_t i=0;i<work_to_do;++i) {
+          container.emplace(i,std::to_string(i));
+      }  
+
+      for (std::size_t i=0;i<work_to_do;++i) {
+          std::string const& item = container[i];
+          if (item != std::to_string(i)) {
+              abort(); // should never get here
```
daniel-j-h: Hm this kills the v8 vm hard, no? Not sure if we should put this into the skeleton code for people to c&p.

src/standalone_async/hello_async.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    bool louder = false;
+
+    // check second argument, should be a 'callback' function.
+    // This allows us to set the callback so we can use it to return errors
+    // instead of throwing as well.
+    if (!info[1]->IsFunction())
+    {
+        Nan::ThrowTypeError("second arg 'callback' must be a function");
+        return;
```
daniel-j-h: We could `return Nan::ThrowTypeError("");` here and below, too.

src/standalone_async/hello_async.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    // Create a worker instance and queues it to run asynchronously invoking the callback when done.
+    // - Nan::AsyncWorker takes a pointer to a Nan::Callback and deletes the pointer automatically.
+    // - Nan::AsyncQueueWorker takes a pointer to a Nan::AsyncWorker and deletes the pointer automatically.
+    auto* worker = new AsyncHelloWorker{louder, new Nan::Callback{callback}};
+    Nan::AsyncQueueWorker(worker);
+
+    // // "info" comes from the NAN_METHOD macro, which returns differently
+    // // according to the version of node
+    // info.GetReturnValue().Set(Nan::New<v8::String>("world").ToLocalChecked());
```
daniel-j-h: I think this block can be remove - we want the function to return nothing and instead have the async worker call the user-provided callback

src/standalone_async/hello_async.cpp
GretaCB: @daniel-j-h great point. I [replaced it with a Callback](https://github.com/mapbox/node-cpp-skel/pull/46/commits/ec533587ef18d79647eb0fa280f4d372b4af923f#diff-881a9b2c4458d0d79b1d20ac96bba548R38), now troubleshooting compile errors.

GretaCB: @daniel-j-h Nevermind, got it going!

GretaCB:
Thanks for your comments @daniel-j-h !!! 🙌

PR #48, opened 2017-06-16
Add object-sync example

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/pull/44, continuing to expand examples. Super rough right now.

- This object-sync example builds off the sync standalone example,  but implements it as part of a `Class`
- Not sure the `hello` framing is very intuitive, so pondering better object use-cases/metaphors
- Gradually removing `src/hello_world.cpp`, and replacing parts of it with this example, and pondering what pieces are best to break apart into other examples

cc @mapbox/core-tech

### DIFF
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -17,6 +17,10 @@ static void init(v8::Local<v8::Object> target) {
     // expose HelloObject class
     object_sync::HelloObject::Init(target);
 
+    // Notice there are multiple "hello" functions as part of this module:
+    // 1) standalone::hello
+    // 2) HelloObject.hello()...not really sure why this function doesnt have to be expliclty exposed
+
     // add more methods/classes below that youd like to use in node.js-world
     // then create a .cpp and .hpp file in /src for each new method
 }
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -16,13 +16,16 @@ static void init(v8::Local<v8::Object> target) {
 
     // expose HelloObject class
     object_sync::HelloObject::Init(target);
+    
+   /**
+    * You may have noticed there are multiple "hello" functions as part of this module.
+    * They are both exposed a bit differently.
+    * 1) standalone::hello    // exposed above
+    * 2) HelloObject.hello    // exposed in object_sync/hello.cpp as part of HelloObject
+    */
 
-    // Notice there are multiple "hello" functions as part of this module:
-    // 1) standalone::hello
-    // 2) HelloObject.hello()...not really sure why this function doesnt have to be expliclty exposed
-
-    // add more methods/classes below that youd like to use in node.js-world
-    // then create a .cpp and .hpp file in /src for each new method
+    // add more methods/classes below that youd like to use in Javascript world
+    // then create a directory in /src with a .cpp and a .hpp file
 }
 
 NODE_MODULE(module, init)
--- a/src/module.cpp
+++ b/src/module.cpp
@@ -5,7 +5,7 @@
 // #include "your_code.hpp"
 
 // "target" is a magic var that nodejs passes into modules scope
-// When you write things to target, they become available to call from js
+// When you write things to target, they become available to call from Javascript world
 static void init(v8::Local<v8::Object> target) {
 
     // expose hello method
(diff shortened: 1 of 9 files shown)


src/module.cpp
```diff
@@ -14,8 +14,12 @@ static void init(v8::Local<v8::Object> target) {
...
     Nan::SetMethod(target, "hello_async", standalone_async::hello_async);
 
-    // expose hello_world class
-    HelloWorld::Init(target);
+    // expose HelloObject class
+    object_sync::HelloObject::Init(target);
+
+    // Notice there are multiple "hello" functions as part of this module:
+    // 1) standalone::hello
+    // 2) HelloObject.hello()...not really sure why this function doesnt have to be expliclty exposed
```
GretaCB: @springmeyer curious why `HelloObject.hello()` doesnt have to be expliclty exposed in the same way as the other methods. Is it because it's handled as part of the class exposure?

src/object_sync/hello.cpp
```diff
@@ -0,0 +1,50 @@
+#include "hello.hpp"
+
+  // If this was not defined within a namespace, it would be in the global scope.
+namespace object_sync {
+  
+  // Default constructor
```
GretaCB: @springmeyer Is this technically a "default constructor"?

src/object_sync/hello.hpp
```diff
@@ -0,0 +1,27 @@
...
+     */
+    class HelloObject: public Nan::ObjectWrap {
+
+        public:
+            // initializer
+            static void Init(v8::Local<v8::Object> target);
+
+            // methods required for the constructor
+            static NAN_METHOD(New);
+            //static Nan::Persistent<v8::Function> &constructor();
```
GretaCB: @springmeyer Any need to add a special constructor for this sync Object example?

Somewhat related, I'm debating what to remove from `src/hello_world.cpp` and add to the new examples.

src/module.cpp
springmeyer: Yes, that is right: the `HelloObject` object/class gets exposed here with `object_sync::HelloObject::Init` and the `HelloObject.hello()` function gets exposed by the `SetPrototypeMethod(fnTp, "hello", hello);` [here](https://github.com/mapbox/node-cpp-skel/pull/48/files#diff-f8bbad9fb9cba460cdc45ebc483664e0R42)

src/object_sync/hello.cpp
springmeyer: Not exactly. The "default constructor" is a C++ term: http://en.cppreference.com/w/cpp/language/default_constructor. In this case the underlying `HelloWorld` C++ is not actually being constructed yet. You are only calling static methods (that are accessible without constructing an instance of the object). However the convention for constructing JS objects is  to call `new SomeObject()` and that JS call will map to this C++ code. So, in a way, this is a default constructor, just not yet in the C++ sense. The default constructor in C++ would be called if you did, in C++ `auto obj = new HellowObject();`.

src/object_sync/hello.hpp
springmeyer: Looks like you are getting away without it, like I mentioned above, you are not actually creating an instance of the C++ HelloWorld class and currently you are only calling static methods.

You'd need to uncomment the `Nan::Persistent<v8::Function> &constructor()` and re-introduce the wrapping at https://github.com/mapbox/node-cpp-skel/blob/7b7e67854fb2a6ea5f95fe70c4cafccf55dc450c/src/hello_world.cpp#L36-L37 if you wanted to store a property on the C++ HelloWorld class and have that accessible in the threadpool. Since this is the primary reason to have a class (to have somewhere to store data) it probably makes sense to re-introduce these things from the `master` branch.

GretaCB:
### Update

Per https://github.com/mapbox/node-cpp-skel/pull/48/files/a55288299aa6393e99257c35556aa38c8ccfdd4d#r123389311, I've started to rethink the design of this object_sync example. I'm running into some issues related to how to access member vars of an instance from within a function, how that plays into the Init function and generally what is happening with the constructor calls. 

I've added a few more comments per chat with @flippmoke , but I'm still a little wobbly with the concepts (ex: Singleton). Would love to run the new design by you, @springmeyer , if you have some time on Monday.

PR #51, opened 2017-07-03
Add build docs

GretaCB:
I think it will be really useful and informative to include docs about what a "build" is, how it is configured, and what the desired conventions we'd like to instill. I've also updated more general docs.

This is super rough, as is my understanding. Would love your 👀  re: the builds docs @flippmoke @springmeyer 

cc @mapsam


README.md
```diff
@@ -90,22 +94,63 @@ module.hello_async({ louder: true }, function(err, result) {
...
+var Obj = new module.HelloObject();
+var hi = Obj.hello();
 
-console.log(hi); // => howdy world!
+console.log(hi); // => "...initialized an object...hello world"
 ```
 
+#### When would you use an object or class?
+- You need to do some preprocessing of data before going into the thread pool. So you're writing the code that makes sense to do that preprocessing once as a separate operation, then going through to the thread pool after the object is ready. Examples:
+  - [vt-shaver](https://github.com/mapbox/vt-shaver-cpp/blob/a5908ccf61c04d54c0b01415ec986d736f5c6c4f/src/filters.hpp#L10): we create the Filter once, then pass the Filter object into Shaver multiple times.
```
GretaCB: remove

Issue #50, opened 2017-06-30
[WIP] Questions/Brainstorming


GretaCB:
Per discussion with @springmeyer , outlining possible next steps for node-cpp-skel and capturing best practices to document and integrate into the skel design.

## Questions to answer
- What does it look like to have the variable that represents an instance of a class accessible in the threadpool?

- What is the role of a singleton and how is it related to other design patterns for sync/async?

- When passing data to thread pool, the data needs to be a type that is safe to pass into the thread pool. What makes a type safe to pass into the thread pool? 
  - Two scenarios that are safe: 
    - Copy data to a C++ object (pure, part of the C++ std lib...not V8)
    - Copy V8 data: Extra work to keep the V8 object alive, avoids the object perhaps getting deleted (trick: use v8::Persistant)

### Goal 1)
- Provide examples of objects that feel real-world, actionable, and sturdy
- Unlocks: ability to fork and run fast and feel confidence the foundation is a best practice

### Goal 2)
- Provide incremental examples that build complexity gradually
- Unlocks: easier for beginning developers to understand gradually why extra complexity is needed

**Preliminary conclusion: object-sync is caught between these worlds**
 
 

## Terms to learn
- [ ] @gretacb to add to cpp glossary

### Standard C++
- Singleton - is a general programming concept for creating a instance once in the process
  - mapnik uses singletons for datasource plugin and font registration
- `static` keyword
- reference (what does the `&` mean?)

### V8
- `Nan::Persistent` (which is really `v8::Persistent`)
- `v8::FunctionTemplate` and `v8::Function`
- `Nan::ObjectWrap`

## Possible next steps:
- [x] @springmeyer could add/edit comments and provide them in another PR
- [x] Build ^^ items into the cpp glossary (@gretacb first pass)
- [x] [Object async example](https://github.com/mapbox/node-cpp-skel/pull/52): Demonstrate how we safely and efficiently utilize the threadpool to do work on data.
  Two options:
  - vt-shaver example:
    - Filters class, exposed in JS
    - Able to create the Filters class once (as object in JS)
    - Standalone `shave` function access Filters as an argument
    - Filters is passed into the threadpool
  - node-mapnik design: Classes expose method

GretaCB:
### Update

Per chat with @springmeyer when running through https://github.com/mapbox/node-cpp-skel/pull/51, I plan to split the README into different docs:
- Reference: Info about C++ best practices, add-on conventions, and general 101
[...]

PR #52, opened 2017-07-11
Add object_async example

GretaCB:
This object-async example builds off the async standalone example and the object-sync example

### Next Actions
- [x] Add code comments and relevant docs ("[When would you use an async object or class?](https://github.com/mapbox/node-cpp-skel/blob/85c81269be34bf28d50c92e8ffe98b4901d699a2/README.md#when-would-you-use-an-object-or-class)")
- [x] All possible Test Coverage for this example
- [x] Add some context around `node::ObjectWrap` and how it relates to `Unwrap`

### Questions
- @springmeyer Does this example hit any of the questions we pondered in https://github.com/mapbox/node-cpp-skel/issues/50 ?
- Does this custom constructor design make sense with an async object? 
- Is this PR relevant to real world usage of an async object?
- Curious exactly what is `node::ObjectWrap` and how it relates to `Unwrap`

Any other thoughts/comments/edits?

cc @springmeyer @daniel-j-h @flippmoke


src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+  *
+  * v8::Local<v8::Function> callback;
+  * CallbackError("error message", callback);
+  * return; // this is important to prevent duplicate callbacks from being fired!
+  */
+  inline void CallbackError(std::string message, v8::Local<v8::Function> callback) {
+  	v8::Local<v8::Value> argv[1] = { Nan::Error(message.c_str()) };
+  	Nan::MakeCallback(Nan::GetCurrentContext()->Global(), callback, 1, argv);
+  }
+
```
springmeyer: I see this same function getting repeated in multiple `.cpp` files. How about we refactor the code to move it to a new header file called `module_utils.hpp`? Then define it there, include `module_utils.hpp` in each of the `cpp` files that uses it, and that will reduce the code duplication.

Note: for fun context because of the `inline` keyword is used the compiler is doing away with this function in the binary code. When something is "inlined" it is re-arranged and optimized. So the `v8::Local<v8::Value>...` and `Nan::MakeCallback...` lines are being moved to where the code is calling `CallbackError`. This means that moving the source code into a common header will only impact how the source code looks. The final binary will probably not change at all because of how the compiler works.

src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+  };  
+
+
+  NAN_METHOD(HelloObjectAsync::hello_async) {
+    // "info" comes from the NAN_METHOD macro, which returns differently
+    // according to the version of node
+    // Mention anything about "Unwrap"?
+    HelloObjectAsync* h = Nan::ObjectWrap::Unwrap<HelloObjectAsync>(info.Holder());
+
+    std::string name = h->name_;
```
springmeyer: Minor issue: this line is resulting in an uncessary copy. We are copying the `h->name_` to a new variable named `name`. Instead, down below you could do:

```
new AsyncHelloWorker{louder, h->name_, new Nan::Callback{callback}};
```

That would avoid the copy. Or alternatively you could do:

```c++
std::string const& name = h->name_;
```

That would also avoid the copy by making the `name` variable a [reference](https://github.com/mapbox/cpp/blob/master/glossary.md#reference) to the `h->name_`.

Why does this matter?

In short, performance. If this code needed to be as fast as possible we want to avoid all possible copies. Because copies require allocation (or re-allocation of memory) for the new variables space. And `std::string` allocates memory on the heap (internally) (https://github.com/mapbox/cpp/blob/master/glossary.md#heap-allocation) when newly created.

src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+  */
+  inline void CallbackError(std::string message, v8::Local<v8::Function> callback) {
+  	v8::Local<v8::Value> argv[1] = { Nan::Error(message.c_str()) };
+  	Nan::MakeCallback(Nan::GetCurrentContext()->Global(), callback, 1, argv);
+  }
+
+
+  // Expensive allocation of std::map, querying, and string comparison,
+  // therefore threads are busy
+  std::string do_expensive_work(bool louder, std::string name) {
```
springmeyer: To avoid a copy here pass by reference like:

```c++
std::string do_expensive_work(bool louder, std::string const& name) {  
```

src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+  }  
+
+  // This is the worker running asynchronously and calling a user-provided callback when done.
+  // Consider storing all C++ objects you need by value or by shared_ptr to keep them alive until done.
+  // Nan AsyncWorker docs: https://github.com/nodejs/nan/blob/master/doc/asyncworker.md
+  struct AsyncHelloWorker : Nan::AsyncWorker {
+
+      using Base = Nan::AsyncWorker;  
+
+      AsyncHelloWorker(bool louder, std::string name, Nan::Callback* callback)
```
springmeyer: Another spot where we can avoid a copy:

Best to pass by reference like:

```c++
AsyncHelloWorker(bool louder, std::string const& name, Nan::Callback* callback)
```

src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+              else
+              {
+                return Nan::ThrowTypeError(
+                    "arg must be a string");
+              }
+            }
+            else {
+                auto *const self = new HelloObjectAsync();
+                self->Wrap(info.This());
+            }
```
springmeyer: I think a more realistic thing would be to throw an error if no argument is passed. Currently we default to `world` below. But I think instead we should throw a JS error here if no `name` arg is passed. What do you think?

springmeyer:
> @springmeyer Does this example hit any of the questions we pondered in #50 ?

Yes, perfectly. We are passing in a custom class that holds a custom data member. That member is safe to pass into the threadpool because it is not a V8 object, but rather a `std::string`.

> Does this custom constructor design make sense with an async object?

You mean the custom constructor that accepts the `name` arg? Yes, I think this makes sense. However I did make one inline comment above that I think we should require this argument rather than allowing the object to be created without an arg.

> Is this PR relevant to real world usage of an async object?

It is. Well done!

> Curious exactly what is node::ObjectWrap and how it relates to Unwrap

We spoke about this voice. The short version is that `node::ObjectWrap` and wrapping/unwrapping objects is the somewhat clumsy way it is possible to bind the two languages. Let's not dig deeper than just knowing:

 - To access a class instance inside a C++ `static method` you must unwrap the object.
 - The C++ methods must be static to make them available at startup across the language boundary.

src/object_async/hello_async.cpp
```diff
@@ -0,0 +1,227 @@
...
+            return Nan::ThrowTypeError(ex.what());
+        }
+
+        info.GetReturnValue().Set(info.This());
+    }
+    else
+    {
+        return Nan::ThrowTypeError(
+            "Cannot call constructor as function, you need to use 'new' keyword");
+    }
```
daniel-j-h: I think we could handle both `new T()` as well as `T()` by adding this to the function's beginning:

```c++
if (!info.IsConstructCall()) {
  auto init = Nan::New(create_once());
  info.GetReturnValue().Set(init->NewInstance());
  return;
}

/*continue with New function impl*/
```

Not sure if we want to do that, though.

If not we could flatten the control flow here by exiting early on `!IsConstructorCall()`.

src/object_async/hello_async.cpp
daniel-j-h: Since you're pointing out string copies here - not sure if it makes sense to use move-semantics throughout this example (e.g. to move the string into the async worker). Maybe it just makes it more complicated than it has to be..

springmeyer: 👍  great suggestion @daniel-j-h - I agree that move semantics here would likely make the most sense given that the goal of node-cpp-skel is to set developers up for writing high performance code.

src/object_async/hello_async.cpp
springmeyer: @daniel-j-h - really great catch 🙇 . I agree that handling the case where a user forgets to use `new` in Javascript is important.  I'm aware that some JS developers like to hide the complexity and allow object creation to work without `new`. But my preference would be to throw since this is a consistent pattern that has worked in other node C++ modules I have written. And I've never seen a complaint about this.

src/object_async/hello_async.cpp
GretaCB: @springmeyer so `inline` is great for reusable methods within a file?

springmeyer: > @springmeyer so inline is great for reusable methods within a file?

- inlining, when in a cpp file, may help performance.
- inlining, when in an hpp file, may help performance and may also be absolutely necessary to allow the header to be included from multiple cpp files when the code in the header includes both the definition and the implementation (which a "header-only" library does). And example of this being fixed is https://github.com/mapbox/vector-tile/pull/19.

So, inlining is important to understand because it impacts performance, code size, and is critical to header-only library design.

> And a util header is one step further than that, great for reusable methods within an application/directory

Exactly. If you want to share code in multiple .cpp files then you need to put at least definitions for it in an .hpp file (and, in the case of header-only design, also the implementation)

Ticketed adding this to the C++ glossary at https://github.com/mapbox/cpp/issues/28

springmeyer: Finished at https://github.com/mapbox/cpp/pull/29

src/object_async/hello_async.cpp
GretaCB: Going to lean towards being more explicit here and requiring `new`

GretaCB:
Thank you so much for your input @daniel-j-h @springmeyer 🙌 

I've updated the PR with the optimization suggestions and comments/links for context. Once tests are 🍏  , I will merge! Feel free to add any other comments if I missed anything.

PR #53, opened 2017-07-25
Reorganize documentation

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/50#issuecomment-314522275, separate reference material from README template.

### Next Actions
- [x] Fill in example code runthrough (`When would you use a standalone function?`)
- [x] Update `API.md` and add links to `Walkthrough example code` sections
- [x] Add link to https://github.com/mapbox/cpp/blob/master/node-cpp.md

cc @springmeyer


src/object_sync/hello.cpp
```diff
@@ -74,15 +71,20 @@ namespace object_sync {
   // NAN_METHOD is applicable to methods you want to expose to JS world
   NAN_METHOD(HelloObject::hello) {
 
-    // Note: a HandleScope is automatically included inside NAN_METHOD (See the docs at NAN that say:
+    // Note: a HandleScope is automatically included inside NAN_METHOD. See the docs at NAN that say:
     // 'Note that an implicit HandleScope is created for you on JavaScript-accessible methods so you do not need to insert one yourself.'
```
GretaCB: @springmeyer Is this comment still relevant? Or does it belong below in Init where the Handlescope is created?

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+- [Software build components](extended-tour.md#software-build-components)
+- [Configuration files](extended-tour.md#configuration-files)
+- [Autogenerated files](extended-tour.md#autogenerated-files)
+- [Overall flow](extended-tour.md#overall-flow)
+
+# Walkthrough example code
+
+This skeleton includes a few examples of how you might design your application, including standalone functions and creating Objects/Classes. Both the synchronous and asynchronous versions of each are included to give an iterative example of what the progression from sync to async looks like. Let's run through reasons why you'd design your code in these ways:
+1. [Standalone function](API.md#hello-1)
+2. [Standalone asynchronous function](../API.md#hello_async)
```
mapsam: This anchor doesn't work - I think it's `#helloasync-1`

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+- [What does "build" mean?](extended-tour.md#what-does-build-mean)
+- [Software build components](extended-tour.md#software-build-components)
+- [Configuration files](extended-tour.md#configuration-files)
+- [Autogenerated files](extended-tour.md#autogenerated-files)
+- [Overall flow](extended-tour.md#overall-flow)
+
+# Walkthrough example code
+
+This skeleton includes a few examples of how you might design your application, including standalone functions and creating Objects/Classes. Both the synchronous and asynchronous versions of each are included to give an iterative example of what the progression from sync to async looks like. Let's run through reasons why you'd design your code in these ways:
+1. [Standalone function](API.md#hello-1)
```
mapsam: This link 404s - add `../` like below should fix it!

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+2. [Standalone asynchronous function](../API.md#hello_async)
+3. [Object/Class](../API.md#helloobject)
+4. [Asynchronous Object/Class](../API.md#helloobjectasync)
+
+
+### When would you use a standalone function?
+Standalone functions are simple and useful when your code is doing some kind of basic operation. They avoid the need for a class, since creating a class in this situation would add unnecessary complexity. Creating an asynchronous standalone function can be powerful if your code has a specific, yet expensive operation to do and would benefit from the use of multiple threads at one time to do that expensive work.
```
mapsam: Maybe give an example of a "basic operation"? Like "is returning the value of two numbers added together" or something? I _think_ vtinfo might be a working example of this? https://github.com/mapbox/vtinfo

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+- **Move semantics**: move semantics avoid data being copied (allocating memory) and are most useful when working with big data. `HelloObjectAsync` is using move semantics to limit unnecessary memory allocation, which can be super expensive and generally is one of the main bottle necks of C++ applications. Other scenarios can also cause bottle necks in your code, but memory allocation is the best-case bottle neck to have because you can either control or avoid it altogether.
+
+Other thoughts related to move semantics:
+- Always best to use move semantics instead of passing by reference, espeically with objects like [`std::string`](http://shaharmike.com/cpp/std-string/), which can be expensive.
+- Relatedly, best to use [`std::unique_ptr`](http://en.cppreference.com/w/cpp/memory/unique_ptr) instead of using `std::shared_ptr` because `std::unique_ptr` is non-copyable. So you're forced to avoid copying, which is a good practice.
+
+
+# Builds
```
mapsam: Really loving this section - thanks!

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Overall flow
+
+The overall flow in terms of software components is:
+
+```
+make -> node-pre-gyp -> node-gyp -> gyp -> make -> compiler/linker
+```
+
+The overall flow, including operations, is:
```
mapsam: 😍

docs/extended-tour.md
GretaCB: Thanks to @springmeyer for writing it 🙌

GretaCB:
Thanks for the review @mapsam !!!

> One final thought I have is it might make sense to move the "XCode" and "Publishing binaries" sections of the README into their own /docs files now that the directory exists. What do you think?

@springmeyer and I were thinkin the README could be a nice template for projects using the skel as a starting point, with general info on how to publish and how to run tests/etc. Does it seem too specific to the skel itself?

src/object_sync/hello.cpp
springmeyer: Yes, still relevant, in the sense that HandleScopes are required whenever you work with V8 objects. Otherwise you will have a memory leak. We work with V8 objects here, so would appear to be leaking memory. However it is not because `NAN_METHOD` takes care of automatically adding a `HandleScope`.

docs/extended-tour.md
springmeyer: How about:

> A standalone function is a function that exists at the top level of the module scope rather than as a member of an object that is instantiated. So if your module is `wonderful` a standalone function would be called like `wonderful.callme()`. A standalone function makes sense when the only data needed by the function can be easily passed as arguments. When it is not easy or clean to pass data as arguments then you should consider encapsulation, an example of which would be exposing a function as a member of an object.

In reality you could achieve clean encapulation either by passing a class/object as an argument to a standalone function (as long as there are public accessor methods to get access to the right data) or by exposing a member function on that class/object that has access to the data. Which is chosen is a matter of preference - both would be equally high performance as long as the class involved pre-calculates/pre-creates as much as possible to be efficient (https://github.com/mapbox/cpp/blob/master/glossary.md#efficiency)

springmeyer:
@GretaCB overall I really love the cleanup and fine-tuning in this PR - great work. I made a few comments where you or @mapsam had questions.

> @springmeyer and I were thinkin the README could be a nice template for projects using the skel as a starting point, with general info on how to publish and how to run tests/etc. Does it seem too specific to the skel itself?

I'm open to any organization, but I actually was thinking that the main readme.md in the root could be a light and easy intro to node-cpp-skel itself. Something that helps orient quickly without too much text. And then the docs folder could contain:

  - more advanced docs on builds, xcode, etc broken out
  - a specific README.template that is what a developer forking node-cpp-skel would use to replace the main README for their module

I'm assuming that audience of skel is developers but the audience of of modules that are published based on skel will be only users: hence why a README.template could be a userful starting point for "hey use this doc that describes the basics for how users will install your module once it is published". Overall however the most important thing is docs on how to use skel I think (rather than the readme.template), as those that fork may take modules a variety of directions..

Thoughts @GretaCB @mapsam?

mapsam:
@GretaCB @springmeyer 👍

GretaCB:
That sounds great @springmeyer , I will reorg a bit, and get a template README up and running within `/docs`

docs/extended-tour.md
```diff
@@ -0,0 +1,159 @@
...
+  - [node-addon-examples](https://github.com/nodejs/node-addon-examples/tree/master/6_object_wrap/nan) for another example of what an object-focused Addon looks like.
+
+Objects/Classes are useful when you start doing more complex operations and need to consider performance more heavily, when performance constraints start to matter. Use classes to compute the value of something once, rather than every time you call a function.
+
+One step further is using an asynchronous object or class, which enables you to pass data into the threadpool in the most efficient way possible. This is the main goal of the `HelloObjectAsync` example in this skel, which also uses "move semantics" for even better performance.
+- **Move semantics**: move semantics avoid data being copied (allocating memory) and are most useful when working with big data. `HelloObjectAsync` is using move semantics to limit unnecessary memory allocation, which can be super expensive and generally is one of the main bottle necks of C++ applications. Other scenarios can also cause bottle necks in your code, but memory allocation is the best-case bottle neck to have because you can either control or avoid it altogether.
+
```
springmeyer: @GretaCB I've added https://github.com/mapbox/cpp/pull/30 which more details on allocation. It is not absolutely correct to say that allocation is super expensive. Really it is quite fast. The issue is that is is required to create objects on the heap. And most programs need to create a lot of objects on the heap. And once all other obvious bottlenecks have been solved, the remaining bottleneck for many programs I've seen is memory allocation. So, reducing allocations is a final stage optimization to get the fastest code possible. Because it is best to design for efficiency, I think avoiding allocations from the start is a great pattern for writing code even when saving those allocations might not produce measurably faster code in isolated benchmarks. But under high load, when many objects in many threads are being allocated - that is the point that the work of reducing unnecessary allocations will really pay off.

Can you remove the `super expensive` reference and link, as you see fit, to the new glossary items on allocation?

docs/extended-tour.md
GretaCB: Thanks @springmeyer , I like using `wonderful.callme()` to demonstrate. Relevant to mention anything about performance benefits by using standalone functions instead of object? For example, I have in my notes `In vt-shaver, we implemented a standalone funciton which is more performant because you dont need to create individual object instances for each use`. Perhaps this over complicates the definition, but seems relevant to the purpose of skel: performance.

springmeyer: > Relevant to mention anything about performance benefits by using standalone functions instead of object

I'm not sure, but great point that this is not totally clear. Can I leave it up to you to figure out how to integrate? I'll offer some more detailed thinking to further flesh out why we designed the way we did in the shaver..... the key to performance is efficiency (so reducing unneeded work - https://github.com/mapbox/cpp/blob/master/glossary.md#efficiency). Therefore, in vt-shaver we:

  - Created a class (`Filter` object) to hold parsed filters because the parsing step (converting a `string` into in-memory C++ objects) took time. We did not want to just pass a `string` object into the `shave` function, so we needed an custom object (a class) to hold the in-memory C++ objects representing the parsed filters in such a way that they could be accessed in a zero-copy way in the threadpool. This optimization worked on the assumption that `Filter` objects could be created once per style and passed per tile when shaving (to avoid needing to parse the filters string per tile). This is what happens in the `shave` function: we pass the `Filters` object and in the threadpool we access its memory to get the actually C++ filters.

We could have added a `shave` function on the `Filters` class. This would have had identical performance to the standalone function we went with. So the reason for the standalone function was, in this case, to separate concerns (https://en.wikipedia.org/wiki/Separation_of_concerns): Filters don't need to know about vector tiles so having a method involving vector tiles on the `Filter` object would have been messy as it combined concerns. They could be separated, so we separated them. So, the `shave` function was standalone because it could be. The key to performance was making the argument a custom class (`Filter` object) to hold pre-processed data needed for shaving.

GretaCB: I can totally see the perhaps trivial but also important difference you're getting at 👍 . I just pushed an edit, based on your thoughts. Thank you!!

docs/extended-tour.md
```diff
@@ -19,7 +19,9 @@ This skeleton includes a few examples of how you might design your application,
...
+A standalone function is a function that exists at the top level of the module scope rather than as a member of an object that is instantiated. So if your module is `wonderful`, a standalone function would be called like `wonderful.callme()`. A standalone function makes sense when the only data needed by the function can be easily passed as arguments. When it is not easy or clean to pass data as arguments then you should consider encapsulation, for example, exposing a function as a member of an object.
+
+One of the benefits of creating a standalone function is that it can help [separate concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). This is mainly a stylistic design decision.
```
springmeyer: 👍

README.md
```diff
@@ -14,8 +14,8 @@ This repository itself can be cloned and edited to your needs. The skeleton prep
 * **Tests**: created with [Tape](https://github.com/substack/tape) in the `test/` directory. Travis CI file is prepared to build and test your project on every push.
 * **Documentation**: use this README as a template and customize for your own project. Also, this skeleton uses [documentation.js](http://documentation.js.org/) to generate API documentation from JSDOC comments in the `.cpp` files. Docs are located in `API.md`.
 * **Build system**: [node-pre-gyp](https://github.com/mapbox/node-pre-gyp) generates binaries with the proper system architecture flags
-* **Publishing**: Structured as a node module with a `package.json` that can be deployed to NPM's registry.
-* **Learning resources**: Read the detailed inline comments within the example code to learn exactly what is happening behind the scenes. Also, check out the [extended tour](./docs/extended-tour.md) to learn more about Node/C++ Addon development, builds, and more specifics about the configuration of this skeleton.
+* **[Publishing](./docs/publishing-binaries.md)**: Structured as a node module with a `package.json` that can be deployed to NPM's registry.
```
springmeyer: @GretaCB - did you forget to commit `./docs/publishing-binaries.md`? (I don't see it in this branch)

springmeyer:
Just did another review. This is looking very close to ready! The only thing I noticed (and commented on above) is that it looks like some files may be missing from `./docs` - perhaps they need to be `git add`'d from your local machine @GretaCB?

docs/extended-tour.md
springmeyer: @GretaCB saw your comment that this was helpful. Great. Also, fyi I just fixed a few typos that were misleading by editing in place.

Issue #54, opened 2017-08-01
Use Nan::Utf8String over String::Utf8Value


springmeyer:
A common way to get data from a v8::String to pass to C++ is:

```c++
std::string str(*v8::String::Utf8Value(info[0].As<String>());
```

While concise (by v8/c++ terms) this can be improved. Instead of `v8::String::Utf8Value` we can use `Nan::Utf8String` and we should validate the input before trying to create a `std::string` by:

  - ensuring the `info[0]` actually is a string
  - ensuring that if it is a string, then the length is > 0 or the string data is null (more on this below)

The `Nan::Utf8String` api both validates strings internally and is fast. It has an [optimized conversion routine](https://github.com/nodejs/nan/blob/7291008d755f658343bcb61f456cb0a6744a6235/nan.h#L907-L924) similar [to node core](https://github.com/nodejs/node/blob/98e54b0bd4854bdb3e2949d1b6b20d6777fb7cde/src/util.cc#L36-L49). And it works across all recent node versions.

So we should convert all usage of `String::Utf8Value` to `Nan::Utf8String`. Like the `String::Utf8Value` interface, the `Nan::Utf8String` API exposes is the ability to check the length of the string before accessing in the value: this is a very good idea for safety and robustness. We don't want to risk passing a null string around, which might create unpredictable behavior or a crash like in https://github.com/mapbox/node-fontnik/issues/124.

To keep 100% test coverage, after adding lines that check the line, we will need to add a test that tries to pass an empty string, and will need to assert reasonable behavior in this case. So, let's move to code like this:

```cpp
  Nan::Utf8String utf8_value(info[0]);
  int len = utf8_value.length();
  if (len <= 0) {
     return Nan::ThrowTypeError("arg must be a non-empty string");
  }
  /// work with string data here....
  // e.g. convert to a std::string
  std::string string_copy(*utf8_value, len);
```

An additional (albeit minor unless the string is massive) performance win of getting the `len` of the string is that we can pass it to the `std::string` constructor. Doing `std::string string_copy(*utf8_value);` would work, but providing the length allows the `std::string` constructor to avoid calculating the length internally and should be faster since it skips an operation. The `std::string` class has a bewildering number of different constructors. This one we are using that accepts a `const char *` as the first arg and an `int` as the second argument is the `(4)` from http://en.cppreference.com/w/cpp/string/basic_string/basic_string.

/cc @GretaCB

GretaCB:
[...]

PR #55, opened 2017-08-04
Nan string

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/54, make sure we use Nan's safer and high performance `Nan::Utf8String` when accepting string args from Javascript world.

cc @springmeyer

### DIFF
--- a/src/object_sync/hello.cpp
+++ b/src/object_sync/hello.cpp
@@ -27,10 +27,10 @@ namespace object_sync {
   // Custom constructor, assigns custom name passed in from Javascript world.
   // This constructor uses member init list via the semicolon, aka "direct initialization" 
   // which is more efficient than using assignment operators.
-  HelloObject::HelloObject(std::string name) : 
-    name_(name) {}
+  HelloObject::HelloObject(std::string && name) : 
+    name_(std::move(name)) {}
 
-  // Triggered from Javascript world when calling "new HelloObjectAsync(name)"
+  // Triggered from Javascript world when calling "new HelloObject(name)"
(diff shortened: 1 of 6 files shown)


src/object_sync/hello.cpp
```diff
@@ -39,7 +39,21 @@ namespace object_sync {
...
+
+                // This line converts a V8 string to a C++ std::string.
+                // In the background, it triggers memory allocation (stack allocating, but std:string is also dynamically allocating memory in the heap)
+                // We want to avoid heap allocation to ensure more performant code.
+                // See https://github.com/mapbox/cpp/blob/master/glossary.md#stack-allocation
+                // and https://stackoverflow.com/questions/79923/what-and-where-are-the-stack-and-heap/80113#80113
+                // Also, providing the length allows the std::string constructor to avoid calculating the length internally 
+                // and should be faster since it skips an operation. 
+                std::string name(*utf8_value, len);
                 auto *const self = new HelloObject(name);
```
springmeyer: Unrelated to your changes (which look great) I noticed that we should be doing:

```
auto *const self = new HelloObject(std::move(name));
```

To avoid copying `name`.

GretaCB: 👍 

@springmeyer are move semantics relevant to synchronous functions/objects as well?

springmeyer: Yes, always relevant when performance matters, so in our case, let's apply wherever possible based on the idea that developers will copy/paste from skel and this code should be fast by default so that best practices trickle into code that eventually will be processing a lot of data, where perf will matter in a realworld way.

Issue #37, opened 2017-02-16
Provide .clang-format and format repository


daniel-j-h:
We should provide a `.clang-format`, format the source code and document how to use it.

springmeyer:
@GretaCB I would recommend tackling this by:

 - [ ] Adding a `.clang-format` in the root directory. For now let's copy https://raw.githubusercontent.com/mapbox/variant/master/.clang-format.
 - [ ] Adding a `./scripts/format.sh` script
   - [ ] Have the script install `clang-format` from mason and automatically format the code in src/*
   - [ ] Have the script return a non-zero exit code if any files have been modified
 - [ ] Adding a `CONTRIBUTING.md` and mention how to run `./scripts/format.sh` (or mention in readme)
 - [ ] Modifying a single travis job to run the `./scripts/format.sh`. If the exit code is non-zero then it will fail, which will remind developers to run the `./scripts/format.sh` locally before pushing to ensure we stay consistent going forward.

Inspired by https://github.com/Project-OSRM/osrm-backend/pull/3513 and https://github.com/Project-OSRM/osrm-backend/pull/3514

springmeyer:
> Adding a ./scripts/format.sh script

Realized this could be tricky and I did not know how to do this myself, so I knocked out a first pass. This uses the existing skel infrastructure for mason:

```bash
#!/usr/bin/env bash

set -eu
set -o pipefail

: '

Runs clang-format on the code in src/

Return `1` if there are files to be formatted, and automatically formats them.

Returns `0` if everything looks properly formatted.

'

./scripts/setup.sh --config local.env
source local.env

mason install clang-format ${MASON_LLVM_RELEASE}
mason link clang-format ${MASON_LLVM_RELEASE}
find src/ -type f -name '*.hpp' -o -name '*.cpp' \
 | xargs -I{} clang-format -i -style=file {}

dirty=$(git ls-files --modified)

if [[ $dirty ]]; then
    echo "The following files have been modified:"
    echo $dirty
    exit 1
else
    exit 0
fi
```

GretaCB:
Sweet, thanks @springmeyer . Looks like your first pass at format.sh did the trick 🎉  over at https://github.com/mapbox/node-cpp-skel/pull/56

PR #56, opened 2017-08-04
Add clang-format

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/37:
- [x] Add clang-format script
- [x] Format changes after running the script
- [x] Add contributing.md with how to run format
- [x] Modifying a single travis job to run the ./scripts/format.sh. If the exit code is non-zero then it will fail, which will remind developers to run the ./scripts/format.sh locally before pushing to ensure we stay consistent going forward.

cc @springmeyer @daniel-j-h


.clang-format
```diff
@@ -0,0 +1,88 @@
...
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        4
+UseTab:          Never
```
daniel-j-h: Flagging here: we got burned a couple of times in osrm because of slight differences in clang-format versions, as in: someone formats with clang-format-3.8 then another dev formats with clang-format-4.0 and we get annoying formatting ping-pong.

What we did was to 1/ use a fixed version for clang-format from mason (this can be different than the clang compiler we're compiling with) and 2/ provide a base line .clang-format config based on that version (`clang-format -style=X -dump-config`) and tune that to your needs.

GretaCB: Thanks for the input @daniel-j-h 🙌

GretaCB: Assuming the format version mirrors the clang compiler version?

springmeyer: Also want to chime in with a thank you @daniel-j-h. I recall that ping-pong issue and that factored into the design of node-cpp-skel, in the sense that we provide a single script `./scripts/setup.sh` to make it easy to use a consistent clang++ version (currently v4.0.0 via https://github.com/mapbox/node-cpp-skel/blob/dbef4683df7d91a261fc7a6a37e81a7aaca6747b/scripts/setup.sh#L7).  The formatting script currently uses this version by requesting clang++ from the `setup.sh` (https://github.com/mapbox/node-cpp-skel/pull/56/commits/dbef4683df7d91a261fc7a6a37e81a7aaca6747b#diff-8cfa73a13763a5d60cbae684e6ba1015R17).

So, I'm imagining all developers will run `./scripts/format.sh` locally and that is also what we'll run remotely on travis. This will ensure we avoid `ping-pong`. It also means that when we upgrade that version we'll have some potential disruption if newer clang++ versions behave differently. If figure this is an acceptable tradeoff.

> 2/ provide a base line .clang-format config based on that version (clang-format -style=X -dump-config) and tune that to your needs.

Great to know this command. I just played around in node-cpp-skel with `./mason_packages/.link/bin/clang-format -style=llvm -dump-config > .clang-format` and seeing the differences. @GretaCB My feeling is that we should not worry about changing the current `.clang-format` you pulled from variant for now. We can entertain adopting a different code style in the future, but this should be discussed more widely at Mapbox. The important thing now is to get the structure in place.

springmeyer: Just checked to see if upgrading to `clang++ 4.0.1` formats the code differently than `clang 4.0.0` given the same `.clang-format`. It does not 🎉 . So, we should plan to upgrade clang++ to the newest from mason (https://github.com/mapbox/mason/pull/452) in another PR. Ticketed this at #58

GretaCB:
Travis is also set 🎉 

screenshot 2017-08-07 14 44 32

.travis.yml
```diff
@@ -113,3 +113,10 @@ matrix:
         - curl -S -f https://codecov.io/bash -o codecov
         - chmod +x codecov
         - ./codecov -x "llvm-cov gcov" -Z
+    # Clang format build
+    - os: linux
+      env: BUILDTYPE=debug
+      node_js: 4
```
springmeyer: This looks good. I feel like an improvement would be to strip way even more details to focus this job exclusively on formatting. It could do `language: generic` and override `install:` to just run the `setup.sh` and not the `make` step - then this would run even faster and would not need the custom `env`.

springmeyer:
One failure on travis unrelated to this PR that happened on OS X:

```
Downloading https://s3.amazonaws.com/mason-binaries/osx-x86_64/clang++/4.0.0.tar.gz
curl: (56) SSLRead() return error -36
bin/clang-4.0: (Empty error message)
tar: Error exit delayed from previous errors.
``

Looks like https://github.com/travis-ci/travis-ci/issues/7636, which appears unsolved, except that some users have switched to wget. No action that we need to take in node-cpp-skel, but mentioning here what I learned nevertheless.

PR #57, opened 2017-08-04
Add example of setting custom error code in C++

springmeyer:
This is not intended to merge, but rather as an example of something to discuss. It shows how to set a custom error code in C++ that can be checked in JS.

Because errors can both happen before entering the threadpool (during the sync code that checks arguments) and inside the threadpool, I needed to apply changes in multiple places. In addition because there is currently no way to trigger an error in the threadpool via JS I added an option that can ask for an error to happen (useful for this demo, not something you'd want in real code).

/cc @mapsam

### DIFF
--- a/src/module_utils.hpp
+++ b/src/module_utils.hpp
@@ -18,8 +18,11 @@ namespace utils {
   * 
   */
   inline void CallbackError(std::string message, v8::Local<v8::Function> callback) {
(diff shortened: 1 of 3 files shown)


springmeyer:
closing now - this served its purpose of being an example, but I don't think is needed to merge.

Issue #63, opened 2017-08-17
Build clang-tidy support into node-cpp-skel


springmeyer:
The `clang-tidy` command is a powerful tool from the llvm suite with the ability to:

 - apply a large set of robustness checks
 - automatically modernize code

Details at https://clang.llvm.org/extra/clang-tidy/

clang-tidy uses [static analysis](https://github.com/mapbox/cpp/blob/master/glossary.md#static-analysis)

We should integrate an easy clang-tidy workflow into node-cpp-skel by following the lead of both nodejs core, mapbox-gl-native, and OSRM:

- https://github.com/nodejs/node/pull/7986
- https://github.com/mapbox/mapbox-gl-native/pull/5325
- https://github.com/Project-OSRM/osrm-backend/commit/998abf05bab47f6d7593e17540fc314c2c294ba9

Note: I presume clang-tidy will find very few issues with the current node-cpp-skel code, since it is fairly simple. The main idea here is to put this in place so that modules based on node-cpp-skel will more readily catch bugs **before** they hit production.

springmeyer:
One challenge here is that clang-tidy needs to compile the code to do its checks. And to compile node-cpp-skel code requires a variety of unique include paths to node-gyp bundled headers and `nan.h` in node_modules, and any mason deps.

To avoid needing to write a second build system to collect and pass all these paths, it is best to have the build system generate a `compile_commands.json` that `clang-tidy` can consume. The format is basically an JSON array of which files are compiled, with what compiler args should be used. It is described in detail at https://clang.llvm.org/docs/JSONCompilationDatabase.html.

The `cmake` build system has support for outputting compile commands this with `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`. But we don't use `cmake` here because the nodejs standard is `gyp`. So the question becomes how to generate the `compile_commands.json` ourselves.

I've found https://github.com/rizsotto/Bear, but it looks like overkill. It would need to be compiled itself, packaged in mason, and who knows how well it works.

So, instead I think we should write a command that captures the output of the existing build, parses it, and rewrites the `compile_commands.json` from it. We can pipe the output of the existing build into this tool in order to capture it.

springmeyer:
Done in #64

Issue #58, opened 2017-08-07
Upgrade to latest llvm + mason


springmeyer:
There is a newer, stable llvm series out (https://github.com/mapbox/mason/pull/452). Not urgent, but also no reason not to upgrade from 4.0.0 to 4.0.1 in the future:

```diff
diff --git a/scripts/setup.sh b/scripts/setup.sh
index 6b9d327..cb41656 100755
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -3,8 +3,8 @@
 set -eu
 set -o pipefail
 
-export MASON_RELEASE="${MASON_RELEASE:-v0.9.0}"
-export MASON_LLVM_RELEASE="${MASON_LLVM_RELEASE:-4.0.0}"
+export MASON_RELEASE="${MASON_RELEASE:-v0.14.1}"
+export MASON_LLVM_RELEASE="${MASON_LLVM_RELEASE:-4.0.1}"
 
 PLATFORM=$(uname | tr A-Z a-z)
 if [[ ${PLATFORM} == 'darwin' ]]; then
```

However, oddly I can't find any llvm 4.0.1 specific release notes yet, only found http://releases.llvm.org/4.0.1/docs/ReleaseNotes.html which look identical to http://releases.llvm.org/4.0.0/docs/ReleaseNotes.html

PR #64, opened 2017-08-17
Implement clang-tidy

springmeyer:
This addresses #63 (see that ticket for context).

This PR builds in:

  - `clang-tidy` checks. If `clang-tidy` detects errors travis will fail
  - The `-fix` command is passed so some fixes are automatically done
  - Other warnings need to be noticed and fixed manually

Other notes:

  - Uses a custom python script to convert the Makefile output into `compile_commands.json`
  - Uses clang-tidy from mason, version controlled in the `setup.sh` script - refs #58

/cc @GretaCB for review

/cc @mapbox/directions @mapbox/gl-core - thank you for the examples of clang-tidy usage in OSRM and MBGL - I learned from them in order to apply here ✊

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,28 @@
+---
+Checks:          '*'
+WarningsAsErrors: '*'
+HeaderFilterRegex: '\/src\/'
+AnalyzeTemporaryDtors: false
+CheckOptions:    
+  - key:             google-readability-braces-around-statements.ShortStatementLines
+    value:           '1'
+  - key:             google-readability-function-size.StatementThreshold
+    value:           '800'
+  - key:             google-readability-namespace-comments.ShortNamespaceLines
+    value:           '10'
+  - key:             google-readability-namespace-comments.SpacesBeforeComments
+    value:           '2'
+  - key:             modernize-loop-convert.MaxCopySize
+    value:           '16'
+  - key:             modernize-loop-convert.MinConfidence
+    value:           reasonable
+  - key:             modernize-loop-convert.NamingStyle
+    value:           CamelCase
+  - key:             modernize-pass-by-value.IncludeStyle
+    value:           llvm
+  - key:             modernize-replace-auto-ptr.IncludeStyle
+    value:           llvm
+  - key:             modernize-use-nullptr.NullMacros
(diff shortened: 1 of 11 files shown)


.clang-tidy
```diff
@@ -0,0 +1,28 @@
+---
+Checks:          '*'
```
springmeyer: Using `*` here is very aggressive. But it does not reveal any more than a few problems inside node-cpp-skel and they are easily fixed. So I think starting with this is ideal. Downstream users of node-cpp-skel will be able to change this to disable some checks (based on https://clang.llvm.org/extra/clang-tidy/) but opt-ing into all of them seems wisest to me to help keep code robust from the start.

.clang-tidy
```diff
@@ -0,0 +1,28 @@
...
+WarningsAsErrors: '*'
```
springmeyer: Added this. However I can't seem to get `run-clang-tidy.py` to return a non-zero error code. So this basically has no impact on the return code, just an minor impact on the stdout from clang-tidy.

.travis.yml
```diff
@@ -123,7 +123,15 @@ matrix:
       env: BUILDTYPE=debug
       node_js: 4
       # Overrides `install` to avoid initializing clang toolchain
-      install: 
-        - ./scripts/format.sh
+      install:
+        # First run the clang-tidy target
+        # This target does not currently return non-zero when problems
+        # are fixed, but rather it fixes automatically. Changed files
+        # will be detected by the format target and will trigger an error
```
springmeyer: Working around the lack of non-zero error code on error mentioned at https://github.com/mapbox/node-cpp-skel/pull/64/files#r133848127 here by running `make tidy` first and depending on `make format` to throw if it sees any changes. It would be ideal to error immediately in `make tidy` but I can't figure out how to get that working...

springmeyer:
🎉  travis properly caught the `clang-tidy` changes to source files and errored: https://travis-ci.org/mapbox/node-cpp-skel/builds/265773011. Digging in the logs for the `make tidy` target show why. Going to now create a separate PR that will address the fixes needed.

springmeyer:
Fixed landed. Will merge this once travis is green.

Issue #65, opened 2017-08-17
clang-tidy error around std::string& member


springmeyer:
While enabling clang-tidy support (#63 and #64) this error came up:

```
../src/object_async/hello_async.cpp:210:5: error: const string& members are dangerous; it is much better to use alternatives, such as pointers or simple constants [google-runtime-member-string-references,-warnings-as-errors]
    std::string const& name_;
    ^
```

And I found: https://clang.llvm.org/extra/clang-tidy/checks/google-runtime-member-string-references.html

This is interesting/surprising. I've used `const std::string &` members before and gotten away clean, I think. The motivation here for @GretaCB and I was to ensure we did not need to copy this string when going into the threadpool. The `name` is owned by the `HelloObjectAsync` class [here](https://github.com/mapbox/node-cpp-skel/blob/d31cd275008c7e109ecb035d0fae2e2236d40cce/src/object_async/hello_async.hpp#L35) and is only temporarily neede by the the `AsyncHelloWorker` that takes the const reference and keeps it as a member [here](https://github.com/mapbox/node-cpp-skel/blob/clang-tidy-fixes/src/object_async/hello_async.cpp#L210).

This warning indicates that member references are worst than a pointer, which would achieve the same purpose. So I'm going to look into using a pointer.

springmeyer:
WOW. Okay, learning a lot here. The docs at https://clang.llvm.org/extra/clang-tidy/checks/google-runtime-member-string-references.html are clear that this is trouble because you could easily, accidentally, send a temporarily into the class and the code would still compile and run. Testing this simulated accident (like at #40) with https://github.com/mapbox/node-cpp-skel/commit/6be05eb56e35d1ce9341438b6fa474ad897d3b87 gives:

|        | OS X/local Release | Asan linux/travis | Linuxrelease                                                   | Linux release                                                  |
|--------|--------------------|-------------------|----------------------------------------------------------------|----------------------------------------------------------------|
| Status | All tests passed 😨   | All tests passed 😨  | Tests failed with segfault or `Error: basic_string::_S_create` ✅ | Tests failed with segfault or `Error: basic_string::_S_create` ✅ |



Results at https://travis-ci.org/mapbox/node-cpp-skel/builds/265783690

Detailed error on linux builds:

```
# success: prints expected string via custom constructor
/home/travis/build/mapbox/node-cpp-skel/test/hello_object_async.test.js:7
    if (err) throw err;
             ^
Error: basic_string::_S_create
[...]

PR #66, opened 2017-08-18
Clang tidy fixes

springmeyer:
Applies fixes to errors that clang-tidy identified - refs #64 / #65

/cc @GretaCB

### DIFF
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -200,7 +200,8 @@ struct AsyncHelloWorker : Nan::AsyncWorker
         v8::Local<v8::Value> argv[argc] = {
             Nan::Null(), Nan::New<v8::String>(result_).ToLocalChecked()};
 
-        callback->Call(argc, argv);
+        // Static cast done here to avoid 'cppcoreguidelines-pro-bounds-array-to-pointer-decay' warning with clang-tidy
+        callback->Call(argc, static_cast<v8::Local<v8::Value>*>(argv));
     }
 
     std::string result_;
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -166,7 +166,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker
 
     using Base = Nan::AsyncWorker;
 
-    AsyncHelloWorker(bool louder, std::string const& name,
+    AsyncHelloWorker(bool louder, const std::string * name,
                      Nan::Callback* callback)
         : Base(callback), result_{""}, louder_{louder}, name_{name} {}
 
@@ -177,7 +177,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker
     {
         try
         {
-            result_ = do_expensive_work(louder_, name_);
+            result_ = do_expensive_work(louder_, *name_);
         }
         catch (const std::exception& e)
         {
@@ -206,9 +206,12 @@ struct AsyncHelloWorker : Nan::AsyncWorker
 
     std::string result_;
     const bool louder_;
-    // We are taking a reference to the name instance passed in from instantiating
-    // HelloObjectAsync above (HelloObjectAsync::New)
-    std::string const& name_;
+    // We use a pointer here to avoid copying the string data.
+    // This works because we know that the original string we are
+    // pointing to will be kept in scope/alive for the time while AsyncHelloWorker
+    // is using it. If we could not guarantee this then we would need to either
+    // copy the string or pass a shared_ptr<std::string>.
+    const std::string * name_;
 };
 
 NAN_METHOD(HelloObjectAsync::helloAsync)
@@ -268,7 +271,7 @@ NAN_METHOD(HelloObjectAsync::helloAsync)
     // - Nan::AsyncQueueWorker takes a pointer to a Nan::AsyncWorker and deletes
     // the pointer automatically.
     auto* worker =
-        new AsyncHelloWorker{louder, h->name_, new Nan::Callback{callback}};
+        new AsyncHelloWorker{louder, &h->name_, new Nan::Callback{callback}};
     Nan::AsyncQueueWorker(worker);
 }
 
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -161,7 +161,7 @@ std::string do_expensive_work(bool louder, std::string const& name)
 // them alive until done.
 // Nan AsyncWorker docs:
 // https://github.com/nodejs/nan/blob/master/doc/asyncworker.md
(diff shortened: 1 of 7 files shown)


src/object_async/hello_async.cpp
```diff
@@ -165,8 +168,11 @@ struct AsyncHelloWorker : Nan::AsyncWorker
 {
 
     using Base = Nan::AsyncWorker;
+    // Make this class noncopyable
+    AsyncHelloWorker(AsyncHelloWorker const&) = delete;
+    AsyncHelloWorker& operator=(AsyncHelloWorker const&) = delete;
 
-    AsyncHelloWorker(bool louder, std::string const& name,
+    AsyncHelloWorker(bool louder, const std::string* name,
```
daniel-j-h: Why by const ptr? Storing a const ref as a member and passing a const ref here does not copy the string. The difference is now the ptr can be null - with references it can not.

springmeyer: > Why by const ptr?

To avoid the possibility of a dangling reference if a temporary string was accidentally passed to `new AsyncHelloWorker{louder, name_copy, new Nan::Callback{callback}};` as simulated at https://github.com/mapbox/node-cpp-skel/commit/6be05eb56e35d1ce9341438b6fa474ad897d3b87. See https://github.com/mapbox/node-cpp-skel/issues/65 for details.

> The difference is now the ptr can be null - with references it can not.

Ugh. Good point. Hard to know the ideal solution here. In general I want to:

a. avoid a copy (allowing both a pointer and a ref do that)
b. do more to make sure the `new AsyncHelloWorker` constructor cannot be misused.

daniel-j-h: Okay I see - if you want to guarantee the string's lifetime is bound to the worker's lifetime we should store a string member by value and in the constructor take a string by value and move it into the member. The user can then decide to either move an argument into the worker's constructor or make a copy (but not using move).

src/object_async/hello_async.cpp
```diff
@@ -222,7 +232,7 @@ NAN_METHOD(HelloObjectAsync::helloAsync)
     // the object.
     // - The C++ methods must be static to make them available at startup across
     // the language boundary (JS <-> C++).
-    HelloObjectAsync* h =
+    auto* h =
```
GretaCB: Why auto here?

src/object_async/hello_async.cpp
```diff
@@ -267,7 +277,7 @@ NAN_METHOD(HelloObjectAsync::helloAsync)
     // - Nan::AsyncQueueWorker takes a pointer to a Nan::AsyncWorker and deletes
     // the pointer automatically.
     auto* worker =
-        new AsyncHelloWorker{louder, h->name_, new Nan::Callback{callback}};
+        new AsyncHelloWorker{louder, &h->name_, new Nan::Callback{callback}};
```
GretaCB: great catch

src/object_async/hello_async.cpp
springmeyer: @daniel-j-h - that is a good idea. Thanks for this back and forth - really helping me realize I need to think this through more. And improving the quality in the process!

In this case my top goal is to avoid a copy. I should spell out a bit more why: it is that I'm imagining this scenario as passing a large amount of data into the threadpool that would be a major perf hit if copied.

So I think the ideal solution here is to 1) rework node-cpp-skel to accept a `node::Buffer` since this would be the more likely the data transport from JS for large data, 2) use a `string_ref` or `string_view` approach to marshal the `node::Buffer` data into the threadpool without copying. This would allow both zero copy access and storing the `string_ref` by value on the `AsyncWorker` so that:

 - we don't need to worry about `clang-tidy` complaining about a `const & ` member
 - we don't need to arbitrarily send down a pointer to avoid the warning
 - we don't hit the -Weffc++11 warning (that happens with a pointer) like `error: ‘struct object_async::AsyncHelloWorker’ has pointer data members [-Werror=effc++] but does not override ‘object_async::AsyncHelloWorker(const object_async::AsyncHelloWorker&)’ [-Werror=effc++]`
 - we are explicit that the data is now owned by the `AsyncHelloWorker` class by using a `string_ref` - that would make it more likely to avoid programming errors with scope going forward.

I've sketched out the background and idea at https://github.com/mapbox/node-cpp-skel/issues/67

src/object_async/hello_async.cpp
springmeyer: @GretaCB - thanks for asking! I'll answer by backing up a bit. A good reason to avoid using `auto` is when using it makes the type hard to quickly understand. Like:

```
auto var = 1.0
auto var2 = var*1;
```

It is not immediately obvious here whether `var2` is a float, double, or integer. This is a case where I don't think `auto` should be used.

But in this case `clang-tidy` automatically fixed the code to use `auto` since it seems like a reasonable use case: it makes the code more concise/shorter but does not obscure the type. Since the type is obvious from the template argument passed to `Unwrap<T>`:

```
auto* h = Nan::ObjectWrap::Unwrap<HelloObjectAsync>(info.Holder());
```

More details at https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html

src/object_async/hello_async.cpp
daniel-j-h: Question re. string_view / string_ref: these abstractions don't own the memory they're pointing to but are basically typed `const&`s - who owns their memory? The buffer? Then who owns the buffer? Do you expect the buffer to be moved into the worker?

springmeyer:
Currently the strong data is owned by the class that calls 'AsyncHelloWorker' so that that data will be certain to stay alive. When we move to using node::Buffer then the memory will be owned by the original Buffer passed in from JS.

src/object_async/hello_async.cpp
springmeyer: The idea currently is that this is safe because the calling class `HelloObjectAsync` owns the `data_` https://github.com/mapbox/node-cpp-skel/blob/e22faa9534e0da61eb3fb5577bbb2976a3a8b09d/src/object_async/hello_async.cpp#L45-L46.  `HelloObjectAsync` will be alive for the whole time its data is used in the threadpool because it the function in JS that is triggering the threadpool usage is a member of `HelloObjectAsync`, the `helloAsync` function.

After this PR, when we move to https://github.com/mapbox/node-cpp-skel/issues/67, then the memory will be owned by the `node::Buffer` and made `Persistant` to ensure it is kept alive.

springmeyer:
This it green, so I'm going to merge. Per https://github.com/mapbox/node-cpp-skel/pull/66#discussion_r135169735, #67 will be a future followup that will enhance some of the code that was changed here. Having `clang-tidy` in place now will help us catch errors sooner in that future work.

PR #77, opened 2017-10-10
More to Attach style brace formatting

springmeyer:
The idea behind providing `clang-format` support in node-cpp-skel is not as much to enforce a single style but more to make it easy for downstream modules to be able to enforce a consistent style. However it became clear when forward porting dawg-cache (https://github.com/mapbox/dawg-cache/pull/23#issuecomment-334534353) that this style for braces is preferable, so I think we should update skel accordingly:

```c++
if (this) {
  ...
}
```

Rather than:

```C++
if (this)
{
 ...
}
```

/cc @apendleton @GretaCB @mapsam

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -33,7 +33,7 @@ BraceWrapping:
   BeforeElse:      true
   IndentBraces:    false
 BreakBeforeBinaryOperators: None
-BreakBeforeBraces: Custom
+BreakBeforeBraces: Attach
 BreakBeforeTernaryOperators: true
 BreakConstructorInitializersBeforeComma: false
 ColumnLimit:     0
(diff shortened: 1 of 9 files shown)


mapsam:
Omg thank you so much 🙏

apendleton:
Yeah to be clear, mostly I wanted the brace style in that specific repo to be consistent with other non-C++ geocoding projects since we move back and forth between JS and C++ projects a lot, and the other style is enforced via lints in our JS code. In general I think there's probably more value in people being able to easily move across projects in different languages within the same team than it is for people to be able to move across projects in the same language across different teams, if that makes sense (like, I'm never going to work on OSRM and I wouldn't expect them to work on carmen-cache, despite them both being written in C++), so I agree that it's not that important for there to be one Mapbox-wide style in C++, especially if there's not one in Javascript.

That said: certainly no objection if you folks prefer it this way as well.

PR #78, opened 2017-10-12
Enable more default warnings

springmeyer:
We should have a more aggressive set of default warnings we enable.

refs mapbox/cpp#37

This brings node-cpp-skel in line with hpp-skel: https://github.com/mapbox/hpp-skel/blob/0af4ae29ec83fe12db20963fcd0855dbe7708a57/CMakeLists.txt#L22

/cc @GretaCB @mapsam

### DIFF
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -154,8 +154,8 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
     AsyncHelloWorker& operator=(AsyncHelloWorker const&) = delete;
 
     AsyncHelloWorker(bool louder, const std::string* name,
-                     Nan::Callback* callback)
-        : Base(callback), result_{""}, louder_{louder}, name_{name} {}
+                     Nan::Callback* cb)
+        : Base(cb), result_{""}, louder_{louder}, name_{name} {}
 
     // The Execute() function is getting called when the worker starts to run.
     // - You only have access to member variables stored in this worker.
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -67,7 +67,7 @@ NAN_METHOD(HelloObjectAsync::New) {
                     * Also, providing the length allows the std::string constructor to avoid calculating the length internally
                     * and should be faster since it skips an operation.
                     */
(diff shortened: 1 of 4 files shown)


springmeyer:
This is green on linux (osx is lagging). going to merge now without OS X.

src/object_async/hello_async.cpp
```diff
@@ -67,7 +67,7 @@ NAN_METHOD(HelloObjectAsync::New) {
                     * Also, providing the length allows the std::string constructor to avoid calculating the length internally
                     * and should be faster since it skips an operation.
                     */
-                    std::string name(*utf8_value, len);
+                    std::string name(*utf8_value, static_cast<std::size_t>(len));
```
mapsam: @springmeyer mind explaining this error a little bit? I'm seeing in vtquery:

```
../src/vtquery.cpp:324:56: error: implicit conversion changes signedness: 'int' to 'size_type' (aka 'unsigned long') [-Werror,-Wsign-conversion]
            std::string geometry(*geometry_utf8_value, geometry_str_len);

```

mapsam: Basically, does the std::string operator expect a size_t type, but will auto convert a number to size_t for us (in this case we don't want these automatic conversions to happen since it's less explicit)?

springmeyer: Yes, `std::string` constructor expects a `std::size_t` which is unsigned. Due to implicit conversion of integer types this will silently convert an `int` (which is signed) into a `std::size_t`, which could be trouble if the `int` is negative (it would flip and be completely the wrong number). So, the warning of `-Wsign-conversion` catches this situation. Now in our case we have a check above for whether the `len` is negative, so we're good. It is therefore safe to suppress the warning by making the conversion from `int` to `std::size_t` explicit with the `static_cast`. So, to recap:

 - implicit conversions are a big source of bugs
 - we want to warn on them always to avoid them from happening
 - we need to protect from the conversions going wrong (numbers wrapping or overflowing)
 - we need to make the conversions safe and explicit.

When these types of conversions are common in code, and handling in a one-off way like we do above is tedious, using something like `numeric_cast` can be helpful. It will convert correctly or thrown when it can't. More on that at https://github.com/mapbox/carmen-cache/issues/94

Issue #81, opened 2017-10-19
Add Code of Conduct


GretaCB:
Per https://blog.mapbox.com/our-code-of-conduct-for-open-source-2b3a81c00c80 , let's start integrating an explicit Code of Conduct to our open source libs.

Here is an [already-existing example](https://blog.mapbox.com/our-code-of-conduct-for-open-source-2b3a81c00c80) from GL-native.

Ideally in the next iteration, we can also include explicit notes on what enforcement and reporting violations looks like (for example, [see ayo.js](https://github.com/ayojs/ayo/blob/latest/CODE_OF_CONDUCT.md#when-something-happens))

cc @mapbox/core-tech

PR #82, opened 2017-10-19
Code of Conduct

GretaCB:
Per https://github.com/mapbox/node-cpp-skel/issues/81

cc @mapbox/core-tech


mapsam:
👍 thank you for adding this @GretaCB!

PR #84, opened 2017-10-30
Add doc for sanitizer flags

GretaCB:
Per chat with @springmeyer .

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,15 @@ Note: by default the build errors on compiler warnings. To disable this do:
 WERROR=false make
 ```
 
+Enable additional sanitizers to catch hard-to-find bugs, for example:
+
+```shell
+export LDFLAGS="-fsanitize=address,undefined,integer"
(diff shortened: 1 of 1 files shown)


springmeyer:
It is a little more involved to get the sanitizers working for node.js projects. This is because we either need to:

 - compile the nodejs binary itself with `-fsanitize=address,undefined,integer`; or
 - we need to preload the address sanitizer shared library (on OS X at least) to get it to work without the nodejs binary being instrumented.

So the blurb you've added will work great only in the case that:

 - A) We're compiling pure C++
 - B) A recent clang++ or g++ is on PATH that supports the sanitizers.

`B` will not be the case on OS X unless we install clang++ via mason or homebrew (because Apple's default clang++ does not support al the sanitizers).

So, all these things combined mean that I think we should provide a custom `make sanitize` command to bake all the puzzle pieces into a single command and hide the complexity.

PR #99, opened 2018-01-03
add npmignore

### LINKED ISSUE #98: Add .npmignore - avoid "invalid ELF header" error
Let's add an .npmignore file to prevent any binaries from being published with the module. Despite the gitignore being present, this can result in, for example, developing on a Mac and accidentally including the mac binaries in the published module - then downstream users installing via linux will hit the following error:

```
Error: /home/travis/...../node_modules/PACKAGE_NAME/lib/binding/PACKAGE.node: invalid ELF header
```

cc @mapbox/core-tech

mapsam:
Resolves #98 

cc @springmeyer @GretaCB

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,9 @@
+lib/binding
+mason_packages
+node_modules
+build
+.toolchain
+.mason
+.git
+.DS_Store
+local.env


springmeyer:
My understanding is that if an `.npmignore` does not exist then the `.gitignore` is used. So, my sense is that an `.npmignore` is only valuable if it differs from the `.gitignore`. In this PR they look the same, right? Are you thinking that it still makes sense to start with an `.npmignore` here because we anticipate with downstream modules the `.npmignore` will start to differ from the `.gitignore`, therefore a `.npmignore` is actually useful?

GretaCB:
Per https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package:
> If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it. 

So does "empty" mean:
A) `.npmignore exists, but no content` ?
B) `.npmignore doesn't exist`? 

Looks like `master` has B, and shouldve reflected the following behaviour based on npm's docs:
> If there's no .npmignore file, but there is a .gitignore file, then npm will ignore the stuff matched by the .gitignore file.

mapsam:
Ah, apparently I had an empty npmignore in vtquery, which led to that scenario. So the absence of one is 👍  but an empty one is 👎 . Happy to close this then!

PR #104, opened 2018-01-18
Fix sanitize testing

springmeyer:
@GretaCB noticed a problem whereby `make sanitize` was only testing one test rather then all of them. The problem was a bug in my original implementation that assumed that `node test/*test.js` would run all the `tape` tests, when in fact we need to do `tape test/*test.js` to run them all (because only `tape` knows how to loop through each). The trick to keep the sanitizers working on OS X (and avoid the problem of losing the `DYLD_INSERT_LIBRARIES` value) is to run tape using node directly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,6 +69,7 @@ matrix:
     - os: linux
       env: BUILDTYPE=debug TOOLSET=asan
       node_js: 4
+      sudo: required
       # Overrides `install` to set up custom asan flags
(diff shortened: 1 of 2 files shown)


GretaCB:
Ah nvm, this was a typo in my local setup. A fresh `fetch` and run of this branch looked gorg 👍  Thanks @springmeyer ! :shipit:

PR #106, opened 2018-02-06
Handle python dep for clang-tidy on OSX

GretaCB:
Due to `run-clang-tidy.py` changing in the latest llvm release per https://reviews.llvm.org/D31326, the `yaml` dependency was added, which is not part of python proper. This causes pip install issues when attempting to install `python-yaml` on OSX.

### Next Actions
- [x] Add workaround in `clang-tidy.sh`
- [x] Add docs

cc @springmeyer

### DIFF
--- a/docs/extended-tour.md
+++ b/docs/extended-tour.md
@@ -206,6 +206,18 @@ make tidy
 - `NOLINT`: In the case that clang-tidy throws for code that you either want to keep as-is or can't control, you can [use `NOLINT`](https://github.com/mapbox/node-cpp-skel/blob/87c8d51f2d7d05ac44f3ea7a607f60e73a2af9c8/src/module.cpp#L43-L46) to silent clang-tidy.
 - Since clang-tidy must compile the code to do its checks, skel [generates](https://github.com/mapbox/node-cpp-skel/blob/master/scripts/generate_compile_commands.py) a JSON file that tells clang-tidy which files to run against and what compile command to run. This newly generated file, `/build/compile_command.json`, is created when running `make tidy`. See [clang's JSON compilation format docs](https://clang.llvm.org/docs/JSONCompilationDatabase.html) for more info.
 
+#### Troubleshooting `clang-tidy`
+`run-clang-tidy.py` changed in the latest llvm release due to https://reviews.llvm.org/D31326, where the yaml dependency was added (which is not part of python proper). On OSX, this requires some steps to install the needed dependency in order to run clang-tidy:
+
+- [ ] Try `pip install pyyaml`
+
+If you get an permissions error or `pip: command not found` error, it's likely you're working with the system Python.
+
+- [ ] We highly recommend installing Python 3 from either Python.org or homebrew.
+This will add a new python binary to your $PATH and allow you to use Python3's builtin `pip3` command in order to install
+without `sudo` and without disrupting the system's Python installation.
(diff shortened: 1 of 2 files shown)


GretaCB:
@springmeyer mind doing a quick gutcheck when you have a chance?

springmeyer:
@GretaCB I have some reservations about this. My perspective is that we should not make decisions about which python version is needed in this script, but rather only a) check to ensure `yaml` is available (what clang-tidy needs) and b) if not available then link to the docs.

springmeyer:
@GretaCB per voice, on my system I've stuck with the system python (version 2) and installed `pip` and `pyyaml`. So I'd love for this PR to keep my setup working AND support the setup of using `python3` from homebrew. So, I've pushed up https://github.com/mapbox/node-cpp-skel/tree/tidy-python-alt. Can you see if that works for you? Its not pretty but should allow both of our setups to work, maybe?

GretaCB:
Hmmm, I seem to be hitting a new error:
```
Traceback (most recent call last):
  File "/Users/carol/dev/node-cpp-skel/mason_packages/.link/share/run-clang-tidy.py", line 42, in <module>
    import Queue
ModuleNotFoundError: No module named 'Queue'
```

Digging in to see where Queue is coming from

GretaCB:
Oh brother, another slight Python2 vs Python3 difference:

```
~/dev/node-cpp-skel??] PRODUCTION ‹$python3
Python 3.6.4 (v3.6.4:d48ecebad5, Dec 18 2017, 21:07:28) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import queue
>>> import Queue
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'Queue'
>>> 
```

```
~/dev/node-cpp-skel??] PRODUCTION ‹$python
Python 2.7.10 (default, Jul 15 2017, 17:16:57) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import queue
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: No module named queue
>>> import Queue
>>> 
```

springmeyer:
woah, that is odd @GretaCB - this is my system:

```
$ python -c "import queue"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named queue
```

```
$ python -c "import Queue;print Queue.__file__"
/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/Queue.pyc
```

GretaCB:
### Next Actions
- [ ] Catch up call with @springmeyer

GretaCB:
Per chat with @springmeyer , we've decided to try vendoring yaml in llvm by adding the following to the `run-clang-tidy.py` file:
```
sys.path.insert(0, os.path.abspath(os.path.dirname(__file__)))
import yaml
```

I'll take a peek at llvm to see if I can figure out how we'd do this, otherwise I'll catch up with @springmeyer in the morning

springmeyer:
Ticketed the plan at https://github.com/mapbox/mason/issues/563, let's chat next actions there.

PR #111, opened 2018-03-06
mason-js port

millzpaugh:
This adds `masonV.0.03-alpha` to npm packages to install C++ packages.


springmeyer:
Note: Let's pause on this one and work to get things solid first over at https://github.com/mapbox/vtquery/pull/67

package.json
```diff
@@ -17,6 +17,7 @@
   "author": "Mapbox",
   "license": "ISC",
   "dependencies": {
+    "mason-js-sdk": "https://github.com/mapbox/mason-js/tarball/fix-linking-a-link",
```
springmeyer: @millzpaugh Could you give a review on https://github.com/mapbox/mason-js/pull/48? If that looks good, would be ideal to publish a new mason-js so this can be updated to use a formal release.

binding.gyp
```diff
@@ -1,5 +1,13 @@
 # This file inherits default targets for Node addons, see https://github.com/nodejs/node-gyp/blob/master/addon.gypi
 {
+  # https://github.com/springmeyer/gyp/blob/master/test/make_global_settings/wrapper/wrapper.gyp
+  'make_global_settings': [
+    ['CXX', '<(module_root_dir)/mason_packages/.link/bin/clang++'],
+    ['CC', '<(module_root_dir)/mason_packages/.link/bin/clang'],
+    ['LINK', '<(module_root_dir)/mason_packages/.link/bin/clang++'],
+    ['AR', '<(module_root_dir)/mason_packages/.link/bin/llvm-ar'],
+    ['NM', '<(module_root_dir)/mason_packages/.link/bin/llvm-nm']
+  ],
```
springmeyer: @GretaCB this is a big change. It moves us to setting the defaults for the compiler toolchain here rather than depending on environment variables only. The motivation was to try to reduce the use of bash as much as possible. This combined with `mason-js` now allows us to use a consistent compiler without needing to use `./scripts/setup.sh` on travis.

package.json
GretaCB: Switch to published version before merging?

scripts/sanitize.sh
```diff
@@ -8,12 +8,39 @@ set -o pipefail
...
+if [[ $(uname -s) == 'Darwin' ]]; then
+    SHARED_LIB_EXT=.dylib
+fi
+
+if [[ $(uname -s) == 'Linux' ]]; then
+    ./node_modules/.bin/mason-js install binutils=2.30 --type=compiled
+    ./node_modules/.bin/mason-js link binutils=2.30 --type=compiled
+    export PATH=$(pwd)/mason_packages/.link/bin:${PATH}
+fi
+export MASON_LLVM_RT_PRELOAD=$(pwd)/$(ls mason_packages/.link/lib/clang/*/lib/*/libclang_rt.asan*${SHARED_LIB_EXT})
```
GretaCB: Awesome 🙌  nice to keep these settings tightly scoped to `sanitize` instead of in the general `setup.sh` file.

package.json
GretaCB: Ah I see, it's still pending https://github.com/mapbox/mason-js/pull/48

binding.gyp
GretaCB: Perhaps we can add something in the docs about how to swap out compilers if someone wanted to use something else and not pull down automatically from mason.

.travis.yml
```diff
@@ -27,71 +23,67 @@ install:
...
+      # Overrides `before_script` (tests are already run in `make sanitize`)
+      before_script:
+
+    ## ** Builds that do not get published **
+    
     # g++ build (default builds all use clang++)
     - os: linux
-      env: BUILDTYPE=debug CXX="g++-6" CC="gcc-6"
-      node_js: 4
+      env: BUILDTYPE=debug CXX="g++-6" CC="gcc-6" LINK="g++-6" AR="ar" NM="nm"
```
GretaCB: Mention this in the docs if someone wanted to use gcc instead?

Makefile
```diff
@@ -1,33 +1,58 @@
+# This Makefile serves a few purposes:
```
GretaCB: Lovely 🙌

Makefile
```diff
@@ -1,33 +1,58 @@
...
 	npm install --ignore-scripts
 
-release: node_modules
+mason_packages/headers: node_modules/nan
+	node_modules/.bin/mason-js install
+
+mason_packages/.link/include: mason_packages/headers
+	node_modules/.bin/mason-js link
+
+build-deps: mason_packages/.link/include
```
GretaCB: Perhaps we can add a code comment here to describe what `build-deps` is since it is now involved in many of the other make commands.

binding.gyp
GretaCB: Also, did a basic google search but couldnt find anything specific. What exactly does `make_global_settings` do? Assuming something like sets `CXX` etc as global env vars that reference these bin commands?

src/module.cpp
```diff
@@ -3,6 +3,7 @@
 #include "standalone/hello.hpp"
 #include "standalone_async/hello_async.hpp"
 #include <nan.h>
+#include <protozero/types.hpp>
```
GretaCB: What necessitated this addition?

springmeyer:
Thanks for the review @GretaCB - next actions on this are:

 - [ ] @millzpaugh updates `mason-js-sdk` version in this PR to latest once it is tagged (we paired today on what will become mason-js `v0.1.3` soon)
 - [ ] @springmeyer makes doc updates based on @GretaCB review (aiming to do this by the end of the week)

scripts/sanitize.sh
springmeyer: Right? This is such a nice cleanup that mason-js inspired/allowed/unlocked.

.travis.yml
springmeyer: Done in https://github.com/mapbox/node-cpp-skel/pull/111/commits/69639045fb7c7f6396de12544e06575805c49900 @GretaCB

Makefile
springmeyer: Started on this in https://github.com/mapbox/node-cpp-skel/pull/111/commits/6979edbf1571433d21880e30f297b3ab281913f6

src/module.cpp
springmeyer: To be able to ensure the mason deps are actually installed (asking the compiler to confirm for us). But, we should remove now that we've tested, since of course `protozero` is not actually a dep here. Removed in ebfd02c

PR #112, opened 2018-03-14
[DONT MERGE] playing around with lib osmium

millzpaugh:
**TL;DR** Write a node-cpp add-on that uses lib-osmium to replace node-osmium in dynoawesome, a tool the cartography team uses to process OSM data. 

**Overview**
Processing OSM data isn’t a format we support in unpacker. node-mapnik is an old project that houses this functionality - we'd ideally like to replace with modular APIs written in C++. Using node-cpp-skel, this replacement would be fairly straight forward. 

Initial steps: 
  - clone node-cpp-skel 
  - lib-osmium as dep with mason-js 
  - write a function that uses lib-osmium to read osm `pbf` files 
  - convert OSM xml file into geojson 
  - output geojson file


springmeyer:
> lib-osmium to replace node-osmium in dynoawesome, a tool the cartography team uses to process OSM data.

Per chat with @rclark - we (at mapbox) are soon no longer going to be using node-osmium, but rather are moving to using `osmium-tool` CLI for our internal data processing needs. So, going to close this usecase that we had in mind when we started this is no longer, and @millzpaugh has since used node-cpp-skel to author https://github.com/mapbox/vtcomposite

PR #116, opened 2018-04-13
Debugging before publish

GretaCB:
Add a command and docs for debugging packages before publishing

@mapsam Any tips/tricks you think are useful per the npm versioning debugging from yesterday?

cc @flippmoke

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 build
 mason_packages
 .DS_Store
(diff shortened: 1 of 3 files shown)


mapsam:
> Any tips/tricks you think are useful per the npm versioning debugging from yesterday?

I still haven't determined the root cause of my publish snafu yesterday ... but once I figure it out I'll report back!

PR #118, opened 2018-04-25
Upgrade to llvm 6 toolchain

springmeyer:
@allieoop packaged the freshly released LLVM 6.0.0 toolchain in mason. This updates node-cpp-skel to start using (and testing) it. If this works then all downstream deps will be encouraged to update too (by forward porting to the latest skel framework).

This change specifically:

 - Uses new packages for llvm 6.0.0 (and sub-packages) mapbox/mason#600 (@allieoop)
 - Uses bundled pyyaml so that should no longer be needed when using >= llvm 6 (refs mapbox/node-cpp-skel#106)

/cc @mapbox/core-tech

### DIFF
--- a/mason-versions.ini
+++ b/mason-versions.ini
@@ -1,8 +1,8 @@
 [headers]
 protozero=1.6.1
 [compiled]
-clang++=5.0.1
-clang-tidy=5.0.1
-clang-format=5.0.1
-llvm-cov=5.0.1
+clang++=6.0.0
+clang-tidy=6.0.0
+clang-format=6.0.0
+llvm-cov=6.0.0
 binutils=2.30


springmeyer:
Looks like this is passing except for the clang-tidy travis job: https://travis-ci.org/mapbox/node-cpp-skel/jobs/370849152#L1190.

It is failing since the clang-tidy job fails if a diff is detected (due to clang-tidy making auto-fixes). This is the diff:

```diff
-        : Base(cb), result_{}, louder_{louder}, name_{name} {}
+        : Base(cb), , louder_{louder}, name_{name} {}
```

Looks like latest clang-tidy has a new feature which was applied: https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html

It is clearly buggy: ` : Base(cb), , louder_{louder}, name_{name} {}` should be ` : Base(cb), louder_{louder}, name_{name} {}`. 

Also, if we make this change it will likely trigger a new warning/error from the `g++ -Weffc++` job due to https://github.com/mapbox/cpp/issues/54

hmmmmm.....

springmeyer:
Interesting. Also hitting another new clang-tidy error due to https://clang.llvm.org/extra/clang-tidy/checks/fuchsia-default-arguments.html. The `fuchsia` term is curious. Looks like a new google operating system: https://en.wikipedia.org/wiki/Google_Fuchsia:

> It first became known to the public when the project appeared on GitHub in August 2016 without any official announcement. It first became known to the public when the project appeared on GitHub in August 2016 without any official announcement. In contrast to prior Google-developed operating systems such as Chrome OS and Android, which are based on Linux kernels, Fuchsia is based on a new microkernel called "Zircon", derived from "Little Kernel",[1][2] a small operating system intended for embedded systems, which was developed by Travis Geiselbrecht, a creator of the NewOS kernel used by Haiku OS.[3] 

Zircon is:

> A subset of the C++14 language is used in the Zircon tree. This includes both the upper layers of the kernel (above the lk layer), as well as some userspace code. In particular, Zircon does not use the C++ standard library, and many language features are not used or allowed.

So, looks like we'll want to ignore all `fuchsia-*` warnings, since node-cpp-skel does not (yet) need to target embedded systems like fuchsia.

springmeyer:
I think this is ready to merge. Going to ask for PR review by @allieoop.

/cc @GretaCB for visibility. 

Here is a rundown of the fixes I made:

> Also, if we make this change it will likely trigger a new warning/error from the g++ -Weffc++ job due to mapbox/cpp#54

Fixed by 7a97785

> So, looks like we'll want to ignore all fuchsia-* warnings

Worked.

> It is clearly buggy: : Base(cb), , louder_{louder}, name_{name} {} should be : Base(cb), louder_{louder}, name_{name} {}.

Filed a bug report against clang-tidy for LLVM 6.0.0: https://bugs.llvm.org/show_bug.cgi?id=37227

.clang-tidy
```diff
@@ -1,28 +1,6 @@
...
-  - key:             modernize-loop-convert.MinConfidence
-    value:           reasonable
-  - key:             modernize-loop-convert.NamingStyle
-    value:           CamelCase
-  - key:             modernize-pass-by-value.IncludeStyle
-    value:           llvm
-  - key:             modernize-replace-auto-ptr.IncludeStyle
-    value:           llvm
-  - key:             modernize-use-nullptr.NullMacros
-    value:           'NULL'
```
springmeyer: Removed this since I don't know its purpose / how it got in there. Thinking it is better to stick with defaults until we have strong opinions on clang-tidy options.

.clang-tidy
```diff
@@ -1,28 +1,6 @@
...
+HeaderFilterRegex: '\/src\/.*'
```
springmeyer: https://github.com/osmcode/libosmium/blob/7674775c11322368a40249eddbc3526eed661c73/.clang-tidy#L89 is my goto for best practices from @joto. I changed this to use `.*` because that clang-tidy does. I've not taken the time to understand the difference/implications between `HeaderFilterRegex: '\/src\/'` and `HeaderFilterRegex: '\/src\/.*'` but we should figure that out...

scripts/clang-tidy.sh
```diff
@@ -30,7 +30,15 @@ if [[ ! -f build/compile_commands.json ]]; then
...
-    make | scripts/generate_compile_commands.py > build/compile_commands.json
+    RESULT=0
+    make > /tmp/make-node-cpp-skel-build-output.txt || RESULT=$?
+    if [[ ${RESULT} != 0 ]]; then
+        echo "Build failed, could not generate compile commands for clang-tidy, aborting!"
+        exit ${RESULT}
+    else
+        cat /tmp/make-node-cpp-skel-build-output.txt | scripts/generate_compile_commands.py > build/compile_commands.json
+    fi
+
```
springmeyer: This change fixes a potential problem that existed before: if the build failed it would leave behind an empty `build/compile_commands.json`. Then the script logic would see that and avoid running the build again, leading clang-tidy to report nothing.

allieoop:
Good find with the new clang-tidy feature bug (https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html). Not great to leave in a comma after removing an argument.

.clang-tidy
allieoop: Seems fine to add these more intentionally as we go, and probably a good idea to document as a reminder as to why we're adding them. 👍

springmeyer:
> Filed a bug report against clang-tidy for LLVM 6.0.0: https://bugs.llvm.org/show_bug.cgi?id=37227

Good news, LLVM dev suggested that this is fixed in `master`. I went and built LLVM head locally with:

```
./mason build llvm 7.0.0
```

Then temporarily modified the `scripts/clang-tidy.sh` to test the 7.0.0 version (which is really HEAD from today) locally:

```dif
diff --git a/scripts/clang-tidy.sh b/scripts/clang-tidy.sh
index 3506d74..d7ff438 100755
--- a/scripts/clang-tidy.sh
+++ b/scripts/clang-tidy.sh
@@ -14,10 +14,10 @@ or errors about things it cannot automatically fix. However I cannot
 figure out how to get this working yet as it seems that clang-tidy
 always returns 0 even on errors.
 '
-
-PATH_TO_CLANG_TIDY_SCRIPT="$(pwd)/mason_packages/.link/share/run-clang-tidy.py"
+BASE=/Users/dane/projects/mason/mason_packages/osx-x86_64/llvm/7.0.0/
+PATH_TO_CLANG_TIDY_SCRIPT="${BASE}/share/clang/run-clang-tidy.py"
 # make sure that run-clang-tidy.py can find the right clang-tidy
-export PATH=$(pwd)/mason_packages/.link/bin:${PATH}
+export PATH=${BASE}/bin:${PATH}
 
 # build the compile_commands.json file if it does not exist
 if [[ ! -f build/compile_commands.json ]]; then
```

and was able to confirm the LLVM bug is solved in latest. Running `make tidy` on `master` of node-cpp-skel no longer results in code that is invalid.

.clang-tidy
springmeyer: > Removed this since I don't know its purpose / how it got in there. 

My guess, however, is that we created with `clang-tidy -dump-config`.

> Seems fine to add these more intentionally as we go, and probably a good idea to document as a reminder as to why we're adding them. 

Great, glad you agree 👍

.clang-tidy
```diff
@@ -1,28 +1,6 @@
...
 ...
```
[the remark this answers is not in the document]
springmeyer: Strangely yes, by measure that `clang-tidy -dump-config` produces it. I don't know its purpose or whether an error would be produced if it were left off.

scripts/clang-tidy.sh
springmeyer: > wondering why you can't just make > build/compile_commands.json? This is probably a sign of my lack of bash knowledge.

The `scripts/generate_compile_commands.py` is needed to reformat the output into this specific format (https://clang.llvm.org/docs/JSONCompilationDatabase.html). Some build systems can do this automatically (like cmake https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) but here we use `gyp`, which cannot. So the `scripts/generate_compile_commands.py` is my attempt to bolt this functionality onto gyp.

> So the second way looks better to me. It's easier to read make > /tmp/make-node-cpp-skel-build-output.txt || RESULT=$? since you just use redirection and || is pretty much universally understood cross languages (I believe).

Great, thanks for the feedback. Its more verbose, but I also find it more readible since it breaks down all the things that are happening (return codes, output, flow) more.

src/standalone_async/hello_async.cpp
```diff
@@ -105,7 +105,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
         callback->Call(argc, static_cast<v8::Local<v8::Value>*>(argv));
     }
 
-    std::string result_;
+    std::string result_{};
```
springmeyer: Note: this `{}` was to appease `-Weffc++` per https://github.com/mapbox/cpp/issues/54

springmeyer:
> So, we will stick with disabling (as needed) in the .clang-tidy config.

Found that disabling is not the only option. It is also pretty easy to avoid these by just being more explicit in the code: 937b762. This has the advantage of not making too many decisions here in skel that might prevent useful warnings downstream. It has the disadvantage of making the code a little more verbose.

PR #128, opened 2018-04-30
Improve the clang-tidy script

springmeyer:
This fixes a longstanding issue in the `scripts/clang-tidy.sh` script. Previously if the build failed it would leave behind an empty `build/compile_commands.json`. Then - when you went to run the `tidy` target again - the script logic would see that `build/compile_commands.json` existed and 
 would avoid running the build again, leading clang-tidy to spuriously report nothing.

This change improves the script to avoid the possibility that it will write an empty `build/compile_commands.json`

Context: The `scripts/generate_compile_commands.py` is needed to reformat the build out into this specific format (https://clang.llvm.org/docs/JSONCompilationDatabase.html). Some build systems can do this automatically (like cmake https://cmake.org/cmake/help/v3.5/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html) but here we use gyp, which cannot. So the `scripts/generate_compile_commands.py` is my attempt to bolt this functionality onto gyp.

/cc @allieoop for final PR. This PR is a subset created via https://github.com/mapbox/node-cpp-skel/pull/118/files#diff-297d66ab0c294f76d1a7ab83006a7383

### DIFF
--- a/scripts/clang-tidy.sh
+++ b/scripts/clang-tidy.sh
@@ -31,7 +31,7 @@ if [[ ! -f build/compile_commands.json ]]; then
     # Run make, pipe the output to the generate_compile_commands.py
     # and drop them in a place that clang-tidy will automatically find them
     RESULT=0
-    make > /tmp/make-node-cpp-skel-build-output.txt || RESULT=$?
+    make | tee /tmp/make-node-cpp-skel-build-output.txt || RESULT=$?
     if [[ ${RESULT} != 0 ]]; then
         echo "Build failed, could not generate compile commands for clang-tidy, aborting!"
         exit ${RESULT}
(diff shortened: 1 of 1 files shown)


springmeyer:
fe692a0 is based on some learning trying to debug hanging clang-tidy jobs in another repo. I think what was happening was:

 - without a command like `tee` the output from `make` was being pushed to a file
 - if `make` took a long time travis would not see any output in `stdout`
 - travis would kill the process since it seemed hung

So, using `tee` ensures that the normal build output is still streamed to travis logs, but also to a file (so that we can then generate the compile_commands.json when needed).

/cc @GretaCB

allieoop:
ah, I see. Thanks for the clear explanation of how `tee` solves this problem. lgtm!

Issue #105, opened 2018-01-27
clang-tidy support followup (actually error travis job with warnings/errors)


springmeyer:
As mentioned at https://github.com/mapbox/node-cpp-skel/pull/64#discussion_r133848127 and https://github.com/mapbox/node-cpp-skel/pull/64#discussion_r133848275 when I set up the `make tidy` target I was unable to find a way to have `clang-tidy` return non-zero on warnings/errors.

What this means is that the `clang-tidy` job on travis will only fail when clang tidy notices problems AND fixes them. Then it will leave behind a git diff and the presence of that diff will trigger the job to fail.

But because clang-tidy cannot auto-fix all problems their is a large class of warnings/errors that do not fail the script and therefore may be present but not failing the travis job.

The todo here is to figure out if the usage of clang-tidy can be fixed (are we doing something wrong, forgetting some flag?) to make `clang-tidy` return a non-zero error code when it encounters warnings/errors.

/cc @GretaCB who saw this recently with shaver - locally @GretaCB saw a number of pretty important clang-tidy warnings that were silently passing on travis since they were not causing the job to fail.

allieoop:
@springmeyer - I believe this is [the bug](https://bugs.llvm.org/show_bug.cgi?id=27628) you are referring to. You can see the [last comment](https://bugs.llvm.org/show_bug.cgi?id=27628#c9) from a couple months ago says "Should be fixed in r329579" where r329579 refers to a code revision which you can see [here](https://reviews.llvm.org/rL329579). The fix is to have LLVM return a non-zero exit code when it sees a compiler error. 

Since this fix is not part of LLVM 6.0.0  we are still coming across this issue. We can test the fix by pointing node-cpp-skel with a dev version of LLVM and its latest changes.

allieoop:
After taking a closer look at the LLVM code change, I see that our issue will not be fixed because we use the `-fix-errors` flag. And you can see here that if `FixErrors` is set, clang-tidy will return zero. You can also take a look at that TODO (my vote is for `b`).

```
  if (FoundErrors) {
    // TODO: Figure out when zero exit code should be used with -fix-errors:
    //   a. when a fix has been applied for an error
    //   b. when a fix has been applied for all errors
    //   c. some other condition.
    // For now always returning zero when -fix-errors is used.
    if (FixErrors)
      return 0;
    if (!Quiet)
      llvm::errs() << "Found compiler error(s).\n";
    return 1;
  }
```

[...]

PR #129, opened 2018-04-30
Upgrade to llvm 6.0.0 redux

springmeyer:
@allieoop packaged the freshly released LLVM 6.0.0 toolchain in mason. This updates `node-cpp-skel` to start using (and testing) it. If this works then all downstream deps will be encouraged to update too (by forward porting to the latest skel framework).

This change specifically:

 - Uses new packages for llvm 6.0.0 (and sub-packages) mapbox/mason#600 (@allieoop)
 - Uses bundled pyyaml such that, when using >= llvm 6, we do not need to manually install an external pyyaml to be able to use run-clang-tidy.py (refs mapbox/node-cpp-skel#106)

Note: this PR replaces https://github.com/mapbox/node-cpp-skel/pull/118

/cc @allieoop

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -69,11 +69,11 @@
       'type': 'loadable_module',
       'dependencies': [ 'action_before_build' ],
       # "make" only watches files specified here, and will sometimes cache these files after the first compile.
-      # This cache can sometimes cause confusing errors when removing/renaming/adding new files. 
+      # This cache can sometimes cause confusing errors when removing/renaming/adding new files.
       # Running "make clean" helps to prevent this "mysterious error by cache" scenario
       # This also is where the benefits of using a "glob" come into play...
       # See: https://github.com/mapbox/node-cpp-skel/pull/44#discussion_r122050205
-      'sources': [ 
+      'sources': [
         './src/module.cpp',
         './src/standalone/hello.cpp',
         './src/standalone_async/hello_async.cpp',
@@ -107,7 +107,7 @@
         'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
         'MACOSX_DEPLOYMENT_TARGET':'10.8',
         'CLANG_CXX_LIBRARY': 'libc++',
-        'CLANG_CXX_LANGUAGE_STANDARD':'c++11',
+        'CLANG_CXX_LANGUAGE_STANDARD':'c++14',
         'GCC_VERSION': 'com.apple.compilers.llvm.clang.1_0'
       }
     }
--- a/common.gypi
+++ b/common.gypi
@@ -2,7 +2,7 @@
   'target_defaults': {
     'default_configuration': 'Release',
     'cflags_cc' : [
-      '-std=c++11',
+      '-std=c++14',
       # The assumption is that projects based on node-cpp-skel will also
       # depend on mason packages. Currently (this will change in future mason versions)
       # mason packages default to being built/linked with the CXX11_ABI=0.
--- a/mason-versions.ini
+++ b/mason-versions.ini
@@ -1,8 +1,8 @@
 [headers]
 protozero=1.6.1
 [compiled]
-clang++=5.0.1
-clang-tidy=5.0.1
-clang-format=5.0.1
-llvm-cov=5.0.1
+clang++=6.0.0
+clang-tidy=6.0.0
+clang-format=6.0.0
+llvm-cov=6.0.0
 binutils=2.30
--- a/package-lock.json
+++ b/package-lock.json
@@ -116,14 +116,6 @@
         "tweetnacl": "^0.14.3"
       }
     },
-    "block-stream": {
-      "version": "0.0.9",
-      "resolved": "https://registry.npmjs.org/block-stream/-/block-stream-0.0.9.tgz",
-      "integrity": "sha1-E+v+d4oDIFz+A3UUgeu0szAMEmo=",
-      "requires": {
-        "inherits": "~2.0.0"
-      }
-    },
     "brace-expansion": {
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
@@ -219,11 +211,6 @@
       "resolved": "https://registry.npmjs.org/deep-equal/-/deep-equal-1.0.1.tgz",
       "integrity": "sha1-9dJgKStmDghO/0zbyfCK0yR0SLU=",
       "dev": true
-    },
-    "deep-extend": {
-      "version": "0.5.1",
-      "resolved": "https://registry.npmjs.org/deep-extend/-/deep-extend-0.5.1.tgz",
-      "integrity": "sha512-N8vBdOa+DF7zkRrDCsaOXoCs/E2fJfx9B9MrKnnSiHNh4ws7eSys6YQE4KvT1cecKmOASYQBhbKjeuDD9lT81w=="
     },
     "define-properties": {
       "version": "1.1.2",
@@ -375,27 +362,6 @@
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8="
-    },
-    "fstream": {
-      "version": "1.0.11",
-      "resolved": "https://registry.npmjs.org/fstream/-/fstream-1.0.11.tgz",
-      "integrity": "sha1-XB+x8RdHcRTwYyoOtLcbPLD9MXE=",
-      "requires": {
-        "graceful-fs": "^4.1.2",
-        "inherits": "~2.0.0",
-        "mkdirp": ">=0.5 0",
-        "rimraf": "2"
-      }
-    },
-    "fstream-ignore": {
-      "version": "1.0.5",
-      "resolved": "https://registry.npmjs.org/fstream-ignore/-/fstream-ignore-1.0.5.tgz",
-      "integrity": "sha1-nDHa40dnAY/h0kmyTa2mfQktoQU=",
-      "requires": {
-        "fstream": "^1.0.0",
-        "inherits": "2",
-        "minimatch": "^3.0.0"
-      }
     },
     "function-bind": {
       "version": "1.1.1",
@@ -601,14 +567,6 @@
       "resolved": "https://registry.npmjs.org/json-schema-traverse/-/json-schema-traverse-0.3.1.tgz",
       "integrity": "sha1-NJptRMU6Ud6JtAgFxdXlm0F9M0A="
     },
-    "json-stable-stringify": {
-      "version": "1.0.1",
-      "resolved": "https://registry.npmjs.org/json-stable-stringify/-/json-stable-stringify-1.0.1.tgz",
-      "integrity": "sha1-mnWdOcXy/1A/1TAGRu1EX4jE+a8=",
-      "requires": {
-        "jsonify": "~0.0.0"
-      }
-    },
     "json-stringify-safe": {
       "version": "5.0.1",
       "resolved": "https://registry.npmjs.org/json-stringify-safe/-/json-stringify-safe-5.0.1.tgz",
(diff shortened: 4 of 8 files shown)


springmeyer:
This is currently blocked by:

```
 ./mason_packages; ./node_modules/.bin/mason-js link
info Symlinked:  /Users/dane/projects/node-cpp-skel/mason_packages/headers/protozero/1.6.1 to  /Users/dane/projects/node-cpp-skel/mason_packages/.link
ERR! Error: EACCES: permission denied, unlink '/Users/dane/projects/node-cpp-skel/mason_packages/.link/include/c++/v1/__bit_reference' 
make[1]: *** [mason_packages/.link] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
```

Which is ticketed at https://github.com/mapbox/mason-js/issues/58. In short, this is not a problem with LLVM 6.0.0, but appears to be a problem in mason-js that surfaces with LLVM 6.0.0.

springmeyer:
Once the above error is solved, we'll hit 2 others related to new features of clang-tidy detecting existing issues in our code:

#### readability-redundant-member-init

These are captured at https://github.com/mapbox/node-cpp-skel/pull/118#issuecomment-384127157 and should be fixable with:

```diff
diff --git a/src/object_async/hello_async.cpp b/src/object_async/hello_async.cpp
index b871a41..22f66bc 100644
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -155,7 +155,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
 
     AsyncHelloWorker(bool louder, const std::string* name,
                      Nan::Callback* cb)
-        : Base(cb), result_{}, louder_{louder}, name_{name} {}
+        : Base(cb), louder_{louder}, name_{name} {}
 
     // The Execute() function is getting called when the worker starts to run.
     // - You only have access to member variables stored in this worker.
@@ -186,7 +186,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
         callback->Call(argc, static_cast<v8::Local<v8::Value>*>(argv));
     }
 
-    std::string result_;
+    std::string result_{};
     const bool louder_;
     // We use a pointer here to avoid copying the string data.
     // This works because we know that the original string we are
diff --git a/src/standalone_async/hello_async.cpp b/src/standalone_async/hello_async.cpp
index 68fb652..72270ee 100644
--- a/src/standalone_async/hello_async.cpp
+++ b/src/standalone_async/hello_async.cpp
@@ -72,7 +72,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
     using Base = Nan::AsyncWorker;
 
     AsyncHelloWorker(bool louder, Nan::Callback* cb)
-        : Base(cb), result_{}, louder_{louder} {}
+        : Base(cb), louder_{louder} {}
 
     // The Execute() function is getting called when the worker starts to run.
     // - You only have access to member variables stored in this worker.
@@ -105,7 +105,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
         callback->Call(argc, static_cast<v8::Local<v8::Value>*>(argv));
     }
 
-    std::string result_;
+    std::string result_{};
     const bool louder_;
 };
 
```

#### fuchsia-default-arguments

These are discussed in https://github.com/mapbox/node-cpp-skel/pull/118#issuecomment-384128216 and should be fixable with:

```diff
diff --git a/src/object_async/hello_async.cpp b/src/object_async/hello_async.cpp
index b871a41..11022f5 100644
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -275,7 +275,7 @@ void HelloObjectAsync::Init(v8::Local<v8::Object> target) {
 
     // Create the HelloObject
     auto fnTp = Nan::New<v8::FunctionTemplate>(
-        HelloObjectAsync::New); // Passing the HelloObject::New method above
+        HelloObject::New, v8::Local<v8::Value>()); // Passing the HelloObject::New method above
     fnTp->InstanceTemplate()->SetInternalFieldCount(
         1);                     // It's 1 when holding the ObjectWrap itself and nothing else
     fnTp->SetClassName(whoami); // Passing the Javascript string object above
diff --git a/src/object_sync/hello.cpp b/src/object_sync/hello.cpp
index f6df549..b9006ed 100644
--- a/src/object_sync/hello.cpp
+++ b/src/object_sync/hello.cpp
@@ -142,7 +142,7 @@ void HelloObject::Init(v8::Local<v8::Object> target) {
 
     // Officially create the HelloObject
     auto fnTp = Nan::New<v8::FunctionTemplate>(
-        HelloObject::New); // Passing the HelloObject::New method above
+        HelloObject::New, v8::Local<v8::Value>()); // Passing the HelloObject::New method above
     fnTp->InstanceTemplate()->SetInternalFieldCount(
         1);                     // It's 1 when holding the ObjectWrap itself and nothing else
     fnTp->SetClassName(whoami); // Passing the Javascript string object above
```

allieoop:
I'm getting that one error where a `make tidy` removes a parameter and leaves the comma. I think we documented this issue somewhere and maybe just manually fixed it the last time it came up.  Do you remember this @springmeyer?

I'll look into how we turn this check off in our `.clang-tidy` file.

```diff
diff --git a/src/object_async/hello_async.cpp b/src/object_async/hello_async.cpp
index b871a41..adbe8ea 100644
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -155,7 +155,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline

     AsyncHelloWorker(bool louder, const std::string* name,
                      Nan::Callback* cb)
-        : Base(cb), result_{}, louder_{louder}, name_{name} {}
+        : Base(cb), , louder_{louder}, name_{name} {}

     // The Execute() function is getting called when the worker starts to run.
     // - You only have access to member variables stored in this worker.
diff --git a/src/standalone_async/hello_async.cpp b/src/standalone_async/hello_async.cpp
index 68fb652..9d09105 100644
--- a/src/standalone_async/hello_async.cpp
+++ b/src/standalone_async/hello_async.cpp
@@ -72,7 +72,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
     using Base = Nan::AsyncWorker;

     AsyncHelloWorker(bool louder, Nan::Callback* cb)
-        : Base(cb), result_{}, louder_{louder} {}
+        : Base(cb), , louder_{louder} {}
```

springmeyer:
> maybe just manually fixed it the last time it came up.

The manual fix I lifted into a diff above at:  https://github.com/mapbox/node-cpp-skel/pull/129#issuecomment-385544411. So I recommend applying that - please go ahead and do that to this branch.

> I think we documented this issue somewhere and 

Not documented anywhere beyond this ticket.

But I did confirm that the problem does not happen with LLVM head (future 7.0.0) on my machine, so I think the manual fix for now is okay since we'll eventually not get bitten by it again. Do you agree? Or maybe we should document this gocha in https://github.com/mapbox/node-cpp-skel/blob/master/docs/extended-tour.md#clang-tidy?

springmeyer:
> But I did confirm that the problem does not happen with LLVM head (future 7.0.0) on my machine

mentioned at https://github.com/mapbox/node-cpp-skel/pull/118#issuecomment-384841898

allieoop:
I agree, manually fixing the clang-tidy change is the right step since we know this won't be an issue later on when we use LLVM 7.0.0. Thanks for confirming!

allieoop:
@springmeyer - I made the change we discussed above and now I think this change is ready to be merged in.

Here's the official 👍

springmeyer:
@allieoop this looks good. One thing is that this branch (I think) is still hitting clang-tidy issues. They are not failing travis due to https://github.com/mapbox/node-cpp-skel/issues/105, but they are visible in the logs: https://travis-ci.org/mapbox/node-cpp-skel/jobs/388585932#L1095 (and if you run locally). I'm 👍 on merging and then ticketing a followup on fixing the clang-tidy issues. What do you think?

allieoop:
Yes, I agree this can be merged as we continue a conversation about #105 (I left some comments there). But before merging I'd like to confirm that:

1) We're both seeing 6 warnings being treated as errors, even though clang-tidy only acknowledges 3 of them.

The 6 errors (originally warnings) that I see are:

hello.cpp
 * error: calling a function that uses a default argument is disallowed
 * error: initializing non-owner 'object_sync::HelloObject *const' with a newly created 'gsl::owner<>'

hello_async.cpp: 
 * error: calling a function that uses a default argument is disallowed
 * error: initializing non-owner 'object_async::HelloObjectAsync *const' with a newly created 'gsl::owner<>'
 * error: initializing non-owner 'object_async::AsyncHelloWorker *' with a newly created 'gsl::owner<>'
 * error: initializing non-owner 'standalone_async::AsyncHelloWorker *' with a newly created 'gsl::owner<>'

Yet clang-tidy prints:
```
3 warnings treated as errors
```

Do you know why only 3 are acknowledged in that statement? To make that there were actually 6 warnings, and not 3 warnings and 3 errors, I removed the line in our .clang-tidy file to treat warnings as errors and all 6 of those errors are indeed originally warnings.

2) The 6 errors are not easy to fix manually.

There are two different types of errors occurring:

 1. Calling a function that uses a default argument, e.g. https://github.com/mapbox/node-cpp-skel/blob/master/src/object_sync/hello.cpp#L144 where we call [Nan::New](https://github.com/nodejs/nan/blob/master/doc/new.md#nannew) but we need to call this function. Since this is part of Nan code and outside of our control I don't see how to fix this error directly.

 2. "Bad" initialization of non-owners with a newly created 'gsl::owner<>', e.g. https://github.com/mapbox/node-cpp-skel/blob/master/src/object_sync/hello.cpp#L72. Looks like this is part of clang-tidy's OwningMemoryCheck: https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/cppcoreguidelines/OwningMemoryCheck.cpp#L316. I found some notes here: https://reviews.llvm.org/D36354 but I need more time to digest this and to figure out how a line of code like this `auto* const self = new HelloObject(std::move(name));` sets it off.

springmeyer:
> Since this is part of Nan code and outside of our control I don't see how to fix this error directly.

I proposed above (see fuchsia-default-arguments) starting to be explicit and not depending on a default arg. That feels like a decent solution to me, you?

> "Bad" initialization of non-owners with a newly created 'gsl::owner<>',

My feeling is that we don't want to use `gls::owner` and I would be up for ignoring this one in our `.clang-tidy` config since this is going to be very difficult to workaround (or at least I think when I've looked before it did not seem possible since we're dealing with the node API here). Maybe when we port to N-API this will go away...

allieoop:
Your fix to the clang-tidy `fuchsia-default-arguments` warning worked. So it turns out that clang-tidy is fine with calling functions that have parameters with default values and only cares whether or not those default values are used. So the guideline is to always pass a value in and never rely on defaults. (This is more a note to myself.)

To address the other warning, `cppcoreguidelines-owning-memory`, which I think is a good [guideline](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines-owning-memory.html#cppcoreguidelines-owning-memory) I switched from using raw pointers to unique_ptrs e.g.:
```diff
-                    auto* const self = new HelloObjectAsync(std::move(name));
+                    const auto self = std::make_unique<HelloObjectAsync>(std::move(name));
```

We have one line of code that creates a raw pointer to a struct which I have clang ignore with the `NOLINT` comment. I do think because I think this check is helpful and worth keeping, but our `AsyncHelloWorker` struct might need some refactoring before it can work with `unique_ptr` which will need to call its constructor, e.g.
```cpp
auto worker = std::make_unique<AsyncHelloWorker>(louder, &h->name_, new Nan::Callback{callback});
```

Update: It looks like Travis isn't finding c++14 which is why tests on Travis are failing so I'll have to into this. @springmeyer maybe you'll see what I'm doing wrong in my attempt to upgrade?

springmeyer:
🙌 Congrats on landing this @allieoop - feels great to be providing more warning free, type safe, and memory-leak protected code code here, thanks to your hard work, methodical debugging, and using the latest LLVM release. The one followup I see is https://github.com/mapbox/node-cpp-skel/issues/141, which I've got on my radar for the coming days.

springmeyer:
note: all the code improvements in this PR we kept but the actual LLVM version was reverted in #146

PR #138, opened 2018-06-06
Fix TOOLSET builds

### LINKED ISSUE #137: Duplicate publishing builds in matrix
It appears https://github.com/mapbox/node-cpp-skel/commit/0ba3b2cf860eeddc36bf84b4e05c2e806e2ed26f#diff-354f30a63fb0907d4ad57269548329e3 may have introduced some duplicate jobs in the travis matrix that are publishable. This results in a race condition when publishing binaries, such that a commit with `[publish binary]` in it will never 100% succeed since node-pre-gyp specifically will not overwrite binaries that already exist on s3.

This was discovered in https://github.com/mapbox/vtcomposite/pull/44

Here is a [node-cpp-skel build](https://travis-ci.org/mapbox/node-cpp-skel) that has some failing jobs since others have already published. The duplicates are across macos and linux. Specifically, here are the two builds that are both producing the following binary:

```
node-cpp-skel/v0.1.0-mapsamtmp/Release/node-v48-linux-x64.tar.gz
```

🍏 [linux job that passed](https://travis-ci.org/mapbox/node-cpp-skel/jobs/388685244)

🍎 [linux job that failed](https://travis-ci.org/mapbox/node-cpp-skel/jobs/388685243) since it produced the same binary but the above job finished first.

Digging a bit further, each of the jobs that result in duplicates are supposed to be "toolset" jobs, which means we pass in something like `TOOLSET=asan` to build a downloadable address sanitizer job. node-pre-gyp is configured to place these in a specific s3 path (defined here with `{toolset}` and defaults to an empty string). We pass the `TOOLSET` matrix variable as an argument down to node-pre-gyp. It expects the argument to be `--toolset` but we are inadvertently [passing it as `--TOOLSET`](https://github.com/mapbox/node-cpp-skel/blob/7ea33af8bc73a7e0fe4a0e3e466fca839ddbb095/.travis.yml#L29) which means it is not picked up. I think if we fix this argument the builds should be working again.

Confirming the toolset builds do not publish as expected, listing out available objects on s3, we would expect two binaries **and** a "toolset" build to exist. node-cpp-skel does not have the toolset.

```
~$ aws s3 ls s3://mapbox-node-binary/@mapbox/node-cpp-skel/v0.1.0-mapsamtmp/Debug/
2018-06-06 11:47:06      43704 node-v48-darwin-x64.tar.gz
2018-06-06 11:46:59     230701 node-v48-linux-x64.tar.gz
```

cc @springmeyer @millzpaugh @GretaCB @allieoop

mapsam:
Resolves #137 by updating the `--TOOLSET` flag to `--toolset` which matches node-pre-gyp's expectations. More background in #137 

Confirmed versions are publishing successfully over at https://travis-ci.org/mapbox/node-cpp-skel/builds/388725890

cc @millzpaugh @springmeyer @GretaCB @allieoop

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,9 @@ before_script:
 
 script:
   # after successful tests, publish binaries if specified in commit message
-  - ./scripts/publish.sh --TOOLSET=${TOOLSET:-} --debug=$([ "${BUILDTYPE}" == 'debug' ] && echo "true" || echo "false")
+  - ./scripts/publish.sh --toolset=${TOOLSET:-} --debug=$([ "${BUILDTYPE}" == 'debug' ] && echo "true" || echo "false")
 
-# the matrix allows you to specify different operating systems and environments to 
+# the matrix allows you to specify different operating systems and environments to
 # run your tests and build binaries
 matrix:
   include:
@@ -79,7 +79,7 @@ matrix:
       before_script:
 
     ## ** Builds that do not get published **
-    
+
     # g++ build (default builds all use clang++)
     - os: linux
       env: BUILDTYPE=debug CXX="g++-6" CC="gcc-6" LINK="g++-6" AR="ar" NM="nm"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mapbox/node-cpp-skel",
-  "version": "0.1.0-mapsamtmp2",
+  "version": "0.1.0",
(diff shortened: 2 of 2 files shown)


springmeyer:
goodness, thanks for catching this subtle cause of failure. I must have done a find and replace that was not case sensitive and should have been. Thank you.

Issue #69, opened 2017-08-25
Enhancement: pass data created in threadpool back to main JS thread in zero-copy way


springmeyer:
The #67 talks about a critical optimization when passing data into the threadpool (to do work on it). The corresponding optimization that is equally important is that large data created in the threadpool needs to be efficiently passed back across the boundary from C++ to JS.

Otherwise if the data is large then this hurts performance in the `After` callback (called `HandleOkCallback` when using `Nan::AsyncWorker`), which runs on the main event loop. In particular, large copies at this point can trigger expensive allocation and further slow down the main event loop.

refs this code: https://github.com/mapbox/node-cpp-skel/blob/44f8d413f58fec46839ed7febb75d70ec1fe584e/src/standalone_async/hello_async.cpp#L104-L113.

## Proposal

Modify node-cpp-skel to return a `node::Buffer` from the threadpool using `node::Buffer::New`.

We'll need to modify the example code to make this logical: we'll need to invent a scenario where a `node::Buffer` would be appropriate. A good example of this is when passing gzip compressed data back from the threadpool.

springmeyer:
> large data created in the threadpool needs to be efficiently passed back across the boundary from C++ to JS

> Modify node-cpp-skel to return a node::Buffer from the threadpool using node::Buffer::New.

The way to achieve both of these things is to allocation data once and move it into the `node.Buffer`. Because the `node.Buffer` object does not support C++ move sematics we can accomplish this with the custom `node::Buffer::New` signature that accepts an existing `char *` to data and also a `free_callback` to use once ownership of that data is passed to `node::Buffer::New`. This can be accomplished like:

- Using a `std::make_unique<std::string>();` to allocate data and work with it (append data, for example) in the threadpool
- The `unique_ptr` will own the data while working on it
- Then exchange ownership with the `node::Buffer` after returning from the threadpool like so:

```
baton->string_ptr.release();
v8::Local<v8::Value> argv[2] = { Nan::Null(),
                                 Nan::NewBuffer((char*)baton->string_ptr->data(),
                                    baton->string_ptr->size(),
                                    [](char *, void * hint) {
                                        delete reinterpret_cast<std::string*>(hint);
                                    },
                                    baton->string_ptr.get()
                                 ).ToLocalChecked()
                               };
Nan::MakeCallback(Nan::GetCurrentContext()->Global(), Nan::New(baton->cb), 2, argv);
```

[...]

PR #147, opened 2018-07-06
Return Buffer object with handoff from unique ptr string

flippmoke:
Added the ability to return the string as a buffer object instead for async methods with this buffer object handing off the result string rather then copying. 

Related issue: #69 

/cc @springmeyer @GretaCB

### DIFF
--- a/src/module_utils.hpp
+++ b/src/module_utils.hpp
@@ -24,4 +24,21 @@ inline void CallbackError(std::string message, v8::Local<v8::Function> func) {
     v8::Local<v8::Value> argv[1] = {Nan::Error(message.c_str())};
     Nan::Call(cb, 1, argv);
 }
+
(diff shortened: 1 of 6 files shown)


flippmoke:
@GretaCB getting `Unknown argument: polyglot` from https://github.com/mapbox/node-cpp-skel/blob/master/package.json#L14

springmeyer:
> @GretaCB getting Unknown argument: polyglot from https://github.com/mapbox/node-cpp-skel/blob/master/package.json#L14

Per chat, this is because you need to install an older `documentation-js` version. @GretaCB can you make sure it is documented how to install the old version? (since I presume it is a big lift to deal with trying to get the newer version working and not something we have time for).

springmeyer:
doc issue ticketed at https://github.com/mapbox/node-cpp-skel/issues/150 and worked around in f9fa032

PR #152, opened 2018-08-20
Upgrade mason deps

springmeyer:
This includes clang++ and clang-tidy packages that both no longer symlink the `c++` directory, which should avoid the (previous) problem with the llvm `6.0.0` packages mentioned at https://github.com/mapbox/node-cpp-skel/pull/146

refs https://github.com/mapbox/mason/pull/650

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,9 +1,9 @@
 ---
-Checks:          '*'
+Checks:          '*,-fuchsia-default-arguments'
 WarningsAsErrors: '*'
 HeaderFilterRegex: '\/src\/'
 AnalyzeTemporaryDtors: false
-CheckOptions:    
+CheckOptions:
   - key:             google-readability-braces-around-statements.ShortStatementLines
     value:           '1'
   - key:             google-readability-function-size.StatementThreshold
@@ -25,4 +25,3 @@ CheckOptions:
   - key:             modernize-use-nullptr.NullMacros
     value:           'NULL'
 ...
-
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -205,7 +205,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
         }
     }
 
-    std::unique_ptr<std::string> result_{};
+    std::unique_ptr<std::string> result_ = std::make_unique<std::string>();
     const bool louder_;
     const bool buffer_;
     // We use a pointer here to avoid copying the string data.
--- a/src/object_async/hello_async.cpp
+++ b/src/object_async/hello_async.cpp
@@ -162,7 +162,6 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
                      const std::string* name,
                      Nan::Callback* cb)
         : Base(cb, "skel:object-async-worker"),
-          result_(),
           louder_(louder),
           buffer_(buffer),
           name_(name) {}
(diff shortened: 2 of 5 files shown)


springmeyer:
@alliecrevier would you be up for giving a quick review? We already did the hard work of upgrading the code to fix all LLVM 6.x clang-tidy issues (as I recall there were many) so this one looks easy. There was just one more place that 6.0.1 caught `readability-redundant-member-init` (and where 6.0.0 must not have been able to). And the `fuchsia-default-arguments` started throwing just with the creation of `std::string(<data>)` which indicates to me that google must not be using libstdc++ (no surprise). But because we do use libtstdc++ and we can easily avoid using `std::string(<data>)`  I think the time has come to just ignore `fuchsia-default-arguments`.

.clang-tidy
```diff
@@ -1,9 +1,9 @@
 ---
-Checks:          '*'
+Checks:          '*,-fuchsia-default-arguments'
```
alliecrevier: So this says that we don't want to enforce using default arguments for declarations or calls. Maybe we should reconsider keeping this check? Having this check would ensure that devs think about what values are used in a default state.

src/object_async/hello_async.cpp
```diff
@@ -206,7 +205,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker // NOLINT to disable cppcoreguideline
         }
     }
 
-    std::unique_ptr<std::string> result_;
+    std::unique_ptr<std::string> result_{};
```
alliecrevier: Can you use `make_unique` instead?

src/standalone_async/hello_async.cpp
```diff
@@ -114,7 +114,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
         }
     }
 
-    std::unique_ptr<std::string> result_;
+    std::unique_ptr<std::string> result_{};
```
alliecrevier: I forget why you have to do this.

.clang-tidy
alliecrevier: Giving this more thought, I think this is a good check to remove. There are cases where having default arguments is helpful.

springmeyer: Yes, the `fuchsia-default-arguments` was triggering this error (just on linux):

```
/home/travis/build/mapbox/node-cpp-skel/build/../src/object_async/hello_async.cpp:249:37: error: calling a function that uses a default argument is disallowed [fuchsia-default-arguments,-warnings-as-errors]
        return utils::CallbackError("first arg 'options' must be an object",
                                    ^
/usr/lib/gcc/x86_64-linux-gnu/4.9/../../../../include/c++/4.9/bits/basic_string.h:494:39: note: default parameter was declared here
      basic_string(const _CharT* __s, const _Alloc& __a = _Alloc());
```

https://travis-ci.org/mapbox/node-cpp-skel/jobs/418057304#L942. Disabling the `fuchsia-default-arguments` fixed the error, which I think is the best route because I can't see an easy way to workaround and I can imagine the usefulness of default args in downstream apps (even if we don't use them in node-cpp-skel).

src/object_async/hello_async.cpp
springmeyer: I wondered about this. I think `std::unique_ptr<std::string> result_{}` may be creating an object that is an invalid/nullptr. Whereas using `make_unique` will create a valid pointer pointing toward an empty string. Let me give that latter a shot since I think that is ideal.

springmeyer: Worked great, thanks for this suggestion @alliecrevier

src/standalone_async/hello_async.cpp
```diff
@@ -114,7 +114,7 @@ struct AsyncHelloWorker : Nan::AsyncWorker {
         }
     }
 
-    std::unique_ptr<std::string> result_{};
+    std::unique_ptr<std::string> result_ = std::make_unique<std::string>();
```
alliecrevier: 👍

alliecrevier:
lgtm!

PR #154, opened 2018-10-29
LLVM 7

springmeyer:
Upgrades to LLVM 7

Refs https://github.com/mapbox/mason/pull/665

### DIFF
--- a/mason-versions.ini
+++ b/mason-versions.ini
@@ -1,8 +1,8 @@
 [headers]
 protozero=1.6.3
 [compiled]
-clang++=6.0.1
-clang-tidy=6.0.1
-clang-format=6.0.1
-llvm-cov=6.0.1
+clang++=7.0.0
+clang-tidy=7.0.0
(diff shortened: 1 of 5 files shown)


springmeyer:
@flippmoke requested your review - would like your quick gut check on whether there is a better way (than NOLINT) to avoid the `bugprone-unused-return-value` warnings, which are new in LLVM 7.

flippmoke:
@springmeyer did a quick review of this all again and I do not think there is a better way. Its just one of those weird gotcha situations where this is actually useful. Not releasing until the `isEmpty()` check makes sure that the constructor for the Buffer actually took ownership of the data from the pointer.

I can think of some other ways to possible do this to make sure that the pointer is always deleted, but then we would have to insert a try/catch here and I am not sure that is ideal either. Overall, I would call this a very valid use of NOLINT.

springmeyer:
Thanks for review @flippmoke

PR #155, opened 2018-11-06
add debug docs for vscode

zmofei:
add how to docs for debugging c++ with the breakpoint in VS Code


zmofei:
cc w/ @springmeyer

springmeyer:
Please feel free to merge @zmofei if you think my edits are okay.

docs/debugging-with-vs-code.md
```diff
@@ -34,15 +42,15 @@ In the `launch.json` you can see some template, like this:
 }
 ```
 
-`launch.json` defined which program you want to run after you click the run debug button, let's say we want run `node test/vtshaver.test.js`(In this command we want to use `node` to run `test/vtshaver.test.js`), so we need to change the configurations, put `program` to node's absolute path(you can use `which node` to find the path of current version of node), and change the `args` to `test/vtshaver.test.js`'s absolute path. 
+`launch.json` defines which program you want to run after you click the run debug button. Let's say we want run `node test/vtshaver.test.js`. In this case we need to change the configurations, put `program` to node's absolute path (you can use `which node` to find the path of current version of node), and change the `args` to `${workspaceFolder}/test/vtshaver.test.js`.
```
zmofei: `args` with `${workspaceFolder}` works for me too, I like this changes 👍

PR #163, opened 2020-04-16
N-api (extra)

artemp:
Use std::string() operator for Napi::String to std::string conversion where appropriate to reduce verbosity but add comment about explicit `Utf8Value()` method

```c++ 
auto str = val.As<Napi::String>().Utf8Value(); //explicit convertion
std::string = val.As<Napi::String>();// operator std::string()

```


springmeyer:
👋🏼 @artemp just noticed this PR. It looks to me like this pre-dates the n-api support landing in master. And so despite it looking like a very big set of changes the only meaningful thing is the last commit (https://github.com/mapbox/node-cpp-skel/pull/163/commits/096ef38aedd4f7026e43bc116cfa5110600e11ab). Is that right? If so, does it still make sense to land the string change improvement? If so, perhaps you could re-create a new PR against latest main branch?

artemp:
> 👋🏼 @artemp just noticed this PR. It looks to me like this pre-dates the n-api support landing in master. And so despite it looking like a very big set of changes the only meaningful thing is the last commit ([096ef38](https://github.com/mapbox/node-cpp-skel/commit/096ef38aedd4f7026e43bc116cfa5110600e11ab)). Is that right? If so, does it still make sense to land the string change improvement? If so, perhaps you could re-create a new PR against latest main branch?

@springmeyer - Looks like https://github.com/mapbox/node-cpp-skel/commit/096ef38aedd4f7026e43bc116cfa5110600e11ab is already in main e.g 

https://github.com/mapbox/node-cpp-skel/blob/main/src/object_async/hello_async.cpp#L155-L159
https://github.com/mapbox/node-cpp-skel/blob/main/src/object_sync/hello.cpp#L24-L28 unless I'm missing something ?

springmeyer:
Ah cool, thanks @artemp - so then this can be closed without any action. Thanks!

